PR #14, opened 2021-06-24
Improvements to composer and Drupal commands

deviantintegral:
This is a bit of a grab-bag of a PR, but the actual number of changed lines is fairly small:

1. Moves `Taskfile.yml` into a subdirectory so we can eventually add one for this project if we want. It's symlinked for now since the installer plugin still looks into the root.
2. Makes it clear the build command is _for production_ use. Otherwise, having a basic "composer install" is not actually useful since we need to run that first to get the taskfile for fresh git checkouts.
3. Switches from checking the existence of generated files to hashing them. Also checks the installed.json file to see if we are including dev dependencies or not.
4. Spells out and namespaces the maintenance mode commands, to make it easier for new users to understand.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1,11 +1,1 @@
-version: '3'
-
-dotenv: ['.env']
-
-includes:
-  drupal: ./vendor/lullabot/drainpipe/tasks/drupal.yml
-
-tasks:
-  default:
-    cmds:
-      - echo "Hello, world!"
+scaffold/Taskfile.yml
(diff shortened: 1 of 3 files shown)


deviantintegral:
Unfortunately, I just discovered that `sources` and `status` do not work together: https://github.com/go-task/task/issues/411

What this means is that if a developer manually runs composer, it's pretty easy to end up in a situation where the current setup won't run the needed composer task.

There's an open, but unreviewed PR up at https://github.com/go-task/task/pull/477

deviantintegral:
The above PR works for me. However, I think what we can do for now is comment out status checking until it's merged.

justafish (approved):

Issue #6, opened 2021-05-19
Create system to define includes / excludes in artifact archives


deviantintegral:
Once a build is complete, we need some way to indicate what files should be included in an archive or not. This could be something with git and gitignores, or another file format that's compatible with some sort of archiver.

deviantintegral:
I've been looking into this. It looks like there may be some limitations in using the gitignore format:

1. Both tar and zip can take a text file with a list of paths to include in an archive. Of course, this could be used with `git add ...` too. I think that's a reasonable base to target - no matter what we use, we want to generate a list of files to include in the archive.
2. Git doesn't have a command that works _absent_ a git index. There's the `git check-index` command, but it explicitly ignores files tracked in the index. That means you can't exclude a file that you want in the repository.
3. Git doesn't have a command or library separated out from git itself. We could possibly write a tool to do so. The closest code is at https://github.com/git/git/blob/master/builtin/check-ignore.c#L71-L125
4. `tar` supports "tcsh globs" for file matching. I'm going to investigate to see if task's sh has support for similar, and if it supports excluding files.

deviantintegral:
I ran into a significant issue with Task; unfortunately, it's shell doesn't support extended globs. It also doesn't support them in sources. Both uses lead to a panic, which can be replicated with the underlying library:

```sh
$ gosh -c 'echo !(c*)'
panic: unhandled word part: *syntax.ExtGlob

goroutine 1 [running]:
mvdan.cc/sh/v3/expand.(*Config).wordFields(0xc0000be140, 0xc0000ce010, 0x1, 0x1, 0xc00001c110, 0x4, 0x0, 0x0, 0x0)
	/Users/andrew/workspace/go/pkg/mod/mvdan.cc/sh/v3@v3.3.0/expand/expand.go:617 +0x1290
mvdan.cc/sh/v3/expand.Fields(0xc0000be140, 0xc00006a430, 0x2, 0x2, 0xc00006a430, 0x1015372, 0xc00006a430, 0xc0000c6ab0, 0x10)
	/Users/andrew/workspace/go/pkg/mod/mvdan.cc/sh/v3@v3.3.0/expand/expand.go:392 +0x158
mvdan.cc/sh/v3/interp.(*Runner).fields(0xc0000c4000, 0xc00006a430, 0x2, 0x2, 0x2, 0x2, 0x0)
[...]

PR #16, opened 2021-06-25
Issue #6: Add a task to create a snapshot of the project and exclude files with .drainpipeignore

justafish:
Another option for #6 (and this is very similar to what we do IRL for pushing to hosts which use git as their deployment strategy)

This leverages the `.gitignore` format, in a similar manner to other projects e.g. [.eslintignore](https://eslint.org/docs/user-guide/configuring/ignoring-code#the-eslintignore-file) and [.dockerignore](https://docs.docker.com/engine/reference/builder/#dockerignore-file)

Example usage: `task release:snapshot -- /tmp/release`

todo: Add a default `.drainpipeignore` file to drainpipe-installer


tasks/release.yml
```diff
@@ -0,0 +1,12 @@
...
+
+tasks:
+  snapshot:
+    desc: Creates a snapshot of the current working directory with files in .drainpipeignore excluded. The target directory should be provided as a cli argument e.g. task release:snapshot -- /tmp/release
+    cmds:
+      - if [ ! -f ".drainpipeignore" ]; then echo ".drainpipeignore does not exist" && exit 1; fi
+      - if [ -d "{{.CLI_ARGS}}" ] || [ "" == "{{.CLI_ARGS}}" ]; then echo "Please provide a path to a directory that does not yet exist" && exit 1; fi
+      - cp -R . {{.CLI_ARGS}}
+      - mv {{.CLI_ARGS}}/.drainpipeignore {{.CLI_ARGS}}/.gitignore
+      - cd {{.CLI_ARGS}} && git add . && git clean -ffdX
```
deviantintegral, line 11: How would I ignore a file that's committed to git, but shouldn't be in build artifacts? For example, developer documentation, or tests?

```sh
ls -la
total 24
drwxr-xr-x   6 andrew  wheel  192 25 Jun 16:17 .
drwxrwxrwt  22 root    wheel  704 25 Jun 16:16 ..
drwxr-xr-x  14 andrew  wheel  448 25 Jun 16:18 .git
-rw-r--r--   1 andrew  wheel    8 25 Jun 16:17 .gitignore
-rw-r--r--   1 andrew  wheel   25 25 Jun 16:17 exclude
-rw-r--r--   1 andrew  wheel   21 25 Jun 16:17 keep

$ git ls-tree HEAD
100644 blob b4317319340befd65497fd2b0494a19687f71e54	exclude
100644 blob ee366ce90075690eeeb1fc9c13416ced3f188916	keep

$ cat .gitignore
exclude

$ ls -la
total 24
drwxr-xr-x   6 andrew  wheel  192 25 Jun 16:17 .
drwxrwxrwt  22 root    wheel  704 25 Jun 16:16 ..
drwxr-xr-x  14 andrew  wheel  448 25 Jun 16:18 .git
-rw-r--r--   1 andrew  wheel    8 25 Jun 16:17 .gitignore
-rw-r--r--   1 andrew  wheel   25 25 Jun 16:17 exclude
-rw-r--r--   1 andrew  wheel   21 25 Jun 16:17 keep
```

Note that "exclude" is still there.

justafish: That should work now

tasks/release.yml
```diff
@@ -0,0 +1,14 @@
+version: '3'
+
+tasks:
+  snapshot:
+    desc: Creates a snapshot of the current working directory with files in .drainpipeignore excluded. The target directory should be provided as a cli argument e.g. task release:snapshot -- /tmp/release
+    cmds:
+      - if [ ! -f ".drainpipeignore" ]; then echo ".drainpipeignore does not exist" && exit 1; fi
+      - if [ -d "{{.CLI_ARGS}}" ] || [ "" == "{{.CLI_ARGS}}" ]; then echo "Please provide a path to a directory that does not yet exist" && exit 1; fi
+      - cp -R . {{.CLI_ARGS}}
```
deviantintegral, line 9: Something to note is that this means that on "live" locals, files are copied to be removed. For example, I have files in `sites/default/files/...`, which total several GB. There's a bit of a penalty copying those to the other directory.

tasks/release.yml
```diff
@@ -0,0 +1,14 @@
...
+      - rm -rf {{.CLI_ARGS}}/.git
+      - cd {{.CLI_ARGS}} && git init
```
deviantintegral, line 11: Is the current directory preserved across multiple lines? I know with other CI tools each line would spawn a fresh shell, unless it was a single string block.

deviantintegral (commented):
A few notes!

tasks/release.yml
deviantintegral: (In particular, I wonder about the `rm` call below)

tasks/release.yml
```diff
@@ -0,0 +1,14 @@
+version: '3'
+
+tasks:
+  snapshot:
+    desc: Creates a snapshot of the current working directory with files in .drainpipeignore excluded. The target directory should be provided as a cli argument e.g. task release:snapshot -- /tmp/release
+    cmds:
+      - if [ ! -f ".drainpipeignore" ]; then echo ".drainpipeignore does not exist" && exit 1; fi
+      - if [ -d "{{.CLI_ARGS}}" ] || [ "" == "{{.CLI_ARGS}}" ]; then echo "Please provide a path to a directory that does not yet exist" && exit 1; fi
```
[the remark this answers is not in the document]
justafish: That will generate a semi-random name which will be hard to capture for use in a later command

tasks/release.yml
```diff
@@ -0,0 +1,14 @@
...
+      - cp -R . {{.CLI_ARGS}}
+      - rm -rf {{.CLI_ARGS}}/.git
```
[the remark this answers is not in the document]
justafish: The `.git` directory is being removed from the destination it was copied to (not the original checkout), so the next line doesn't fail

tasks/release.yml
justafish: It isn't, though you can preserve it with a pipe I believe

justafish:
@deviantintegral I've made some updates and added another task here along the same lines, but exports as an archive

tasks/snapshot.yml
```diff
@@ -0,0 +1,35 @@
...
+        cd $TMP_DIR
+        rm -rf .git
+        git init
+        mv .drainpipeignore .gitignore
+        git add .
+        git clean -ffdX
+        git config user.email 'no-reply@lullabot.com'
+        git config user.name 'Lullabot Drainpipe'
+        git commit -m "Initial commit" --no-gpg-sign
+        git archive --format={{.format}} HEAD
```
deviantintegral, line 35: From `git archive --help`:

> git archive behaves differently when given a tree ID versus when given
>        a commit ID or tag ID. In the first case the current time is used as
>        the modification time of each file in the archive. In the latter case
>        the commit time as recorded in the referenced commit object is used
>        instead.

Does that mean we're losing modification timestamps since we're creating a whole new repository?

In `ATTRIBUTES export-ignore`: does that mean that `.gitattributes` could also change what files are included / excluded? It seems like a possible point of confusion that'd be tricky to debug.

deviantintegral (commented):
Notes from reading - I haven't actually had a chance to try this out yet.

scaffold/Taskfile.yml
```diff
@@ -4,3 +4,4 @@ dotenv: ['.env']
 
 includes:
   drupal: ./vendor/lullabot/drainpipe/tasks/drupal.yml
+  snapshot: ./vendor/lullabot/drainpipe/tasks/snapshot.yml
```
deviantintegral, line 7: A next-day thought; is "artifact" or "archive" a more common term than "snapshot"? Especially as we use snapshot in the context of "development snapshot".

tasks/snapshot.yml
```diff
@@ -0,0 +1,35 @@
+version: '3'
+
+tasks:
+  directory:
+    desc: "Creates a snapshot of the current working directory with .git, files in .drainpipeignore, and .drainpipeignore excluded. The target directory should be provided as a cli argument e.g. task snapshot:directory -- /tmp/release"
```
deviantintegral, line 5: Another question about Task; how does it handle per-task help? I expected to be able to do something like `task <command> --help`. It looks like it can only do a single line description as a part of `task --list`, which means we can't just stuff a long description there.

deviantintegral (changes requested):
I've given this a run, and I think I may have found a blocker with using git for this :(

gitignore parsing also includes _global_ exclude rules. For example, my testing ignore file only contained `.vagrant` (a good test, as it has a big vdi in it!). Note how the archive excluded other files:

```
warning: adding embedded git repository: web/modules/contrib/views_tree
warning: adding embedded git repository: web/themes/contrib/olivero
Removing .DS_Store
Removing .idea/
Removing .vagrant/
Removing private/files/.DS_Store
Removing private/files/pictures/.DS_Store
Removing private/files/styles/.DS_Store
Removing private/files/styles/medium/.DS_Store
Removing private/files/styles/medium/private/.DS_Store
Removing private/files/styles/thumbnail/.DS_Store
Removing private/files/styles/thumbnail/private/.DS_Store
Removing vendor/grasmash/yaml-expander/build/
Removing web/core/modules/system/tests/fixtures/HtaccessTest/access_test.module.swp
Removing web/core/modules/system/tests/fixtures/HtaccessTest/access_test.php.swp
Removing web/modules/contrib/type_tray/assets/.DS_Store
Removing web/sites/default/files/.DS_Store
Removing web/sites/default/files/media-icons/.DS_Store
Removing web/sites/default/files/styles/.DS_Store
Removing web/sites/default/files/styles/thumbnail/.DS_Store
Removing web/sites/default/files/styles/thumbnail/public/.DS_Store
Removing web/sites/default/files/styles/thumbnail/public/media-icons/.DS_Store
```

`man gitignore` details all the different files and variables that are used in calculating the final list of patterns.

It seems to me that building an archive should be as deterministic as we can make it, and that having out-of-working-tree files used for the build is likely to lead to confusion down the line. Thoughts?

tasks/snapshot.yml
justafish: https://taskfile.dev/#/usage?id=display-summary-of-task

tasks/snapshot.yml
```diff
@@ -0,0 +1,35 @@
...
+      - |
+        cd {{.CLI_ARGS}}
+        rm -rf .git
+        git init
+        mv .drainpipeignore .gitignore
+        git add .
+        git clean -ffdX
+        rm -rf .git
+  archive:
+    silent: true
```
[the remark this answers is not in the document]
justafish: I think it's useful to see the output of running steps - just not here because the output is piped elsewhere :smile:

tasks/snapshot.yml
```diff
@@ -0,0 +1,35 @@
+version: '3'
+
+tasks:
+  directory:
+    desc: "Creates a snapshot of the current working directory with .git, files in .drainpipeignore, and .drainpipeignore excluded. The target directory should be provided as a cli argument e.g. task snapshot:directory -- /tmp/release"
+    cmds:
+      - if [ -d "{{.CLI_ARGS}}" ] || [ "" == "{{.CLI_ARGS}}" ]; then echo "Please provide a path to a directory that does not yet exist" && exit 1; fi
+      - cp -R . {{.CLI_ARGS}}
+      - if [ ! -f ".drainpipeignore" ]; then echo ".drainpipeignore does not exist" && touch {{.CLI_ARGS}}/.drainpipeignore; fi
```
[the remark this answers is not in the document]
justafish: projects don't need that file - the file is moved to `.gitignore` later on, the empty file is just so that operation doesn't fail.

tasks/snapshot.yml
```diff
@@ -0,0 +1,35 @@
...
+        mv .drainpipeignore .gitignore
+        git add .
+        git clean -ffdX
+        rm -rf .git
+  archive:
+    silent: true
+    desc: "Creates a snapshot of the current working directory as an archive. Format options are tar, tar.gz, or zip. Usage: task snapshot:archive format=zip > ~/archive.zip"
+    cmds:
+      - |
+        TMP_DIR=$(mktemp -d)
```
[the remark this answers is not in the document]
justafish: I've moved it to a variable so that it doesn't appear as an available task

tasks/snapshot.yml
justafish: For the directory export they are preserved, for the archive export no. Given that all of the hosting providers we work with are using the directory export/git remote, and the timestamp of a file is inherently a local attribute I don't think this is an issue? Happy to switch the archive command out if you've got an example otherwise though :+1:

I've documented the behaviour with .gitattributes - that's actually kind of cool and potentially useful!

justafish:
@deviantintegral the global gitignore issue should be resolved

deviantintegral (changes requested):
Still adding additional ignores, but not as many?

```
Initialized empty Git repository in /private/tmp/snapshot/.git/
warning: adding embedded git repository: web/modules/contrib/views_bulk_edit
hint: You've added another git repository inside your current repository.
hint: Clones of the outer repository will not contain the contents of
hint: the embedded repository and will not know how to obtain it.
hint: If you meant to add a submodule, use:
hint:
hint: 	git submodule add <url> web/modules/contrib/views_bulk_edit
hint:
hint: If you added this path by mistake, you can remove it from the
hint: index with:
hint:
hint: 	git rm --cached web/modules/contrib/views_bulk_edit
hint:
hint: See "git help submodule" for more information.
warning: adding embedded git repository: web/modules/contrib/views_tree
warning: adding embedded git repository: web/themes/contrib/olivero
Removing .idea/workspace.xml
Removing .vagrant/
Removing vendor/grasmash/yaml-expander/build/
❯ cat .drainpipeignore
.vagrant
```

Aside, in double checking I noticed "Patterns read from a .gitignore file in the same directory as the path, or in any parent directory, ". It doesn't look like you can turn off the parent directory parsing?

I reran this commenting out the final rm's and didn't see anything odd in repo. For testing, I added `composer.json` to my ignore file in `~/.git/config/ignore` and rebuilt.

Ahah! Git is also parsing `.gitignore` _down_ the directory tree too. For example, a D8 install has `./vendor/grasmash/yaml-expander/.gitignore`, which causes the build directory that is shipped with composer to be removed.

justafish:
I'm not sure what you mean with the parent directory parsing - could you provide an example? I've never heard of a .gitignore in a parent directory of the repository root having an affect (other than the global gitignore).

I borrowed some lines from our previous project to clean out .gitignore and .git - could you test again please? :pray:

tasks/snapshot.yml
```diff
@@ -2,7 +2,8 @@ version: '3'
 
 vars:
   PREPARE_DIRECTORY: |
-    rm -rf .git
+    find . -type d -name ".git" | xargs rm -rf
```
deviantintegral, line 5: I think we probably need to use `-print0` for find and `-0` for xargs, in case there's any spaces in file or directory names?

deviantintegral:
> I'm not sure what you mean with the parent directory parsing - could you provide an example?

Let me try, I was just reading the man page for what it says!

```
❯ git status
On branch main

No commits yet

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	ignored-from-outer-directory

nothing added to commit but untracked files present (use "git add" to track)
❯ cat ../.gitignore
ignored-from-outer-directory
```

Ok, well that's good. I actually think the docs could use an update, and "Patterns read from a .gitignore file in the same directory as the path, or in any parent directory" should be "Patterns read from a .gitignore file in the same directory as the path, or in any parent directory **in the repository**".

tasks/snapshot.yml
justafish: fixed :+1:

deviantintegral (approved):
I gave this a thorough testing, and excludes seem to be working well! 👏 

In testing against a fresh Drupal 9 composer project, I noticed the following warnings:

```
task: [snapshot:directory] if [ -d "/tmp/release2" ] || [ "" == "/tmp/release2" ]; then echo "Please provide a path to a directory that does not yet exist" && exit 1; fi
task: [snapshot:directory] cp -pR . /tmp/release2
task: [snapshot:directory] if [ ! -f ".drainpipeignore" ]; then echo ".drainpipeignore does not exist" && touch /tmp/release2/.drainpipeignore; fi
.drainpipeignore does not exist
task: [snapshot:directory] cd /tmp/release2
find . -type d -name ".git" -print0 | xargs -0 rm -rf
find . -name ".gitignore" -print0 | xargs -0 rm -rf
git init
mv .drainpipeignore .gitignore
# Override global gitignore
git config core.excludesFile .gitignore
git config core.attributesfile .gitattributes
git add .
git clean -ffdX
rm .gitignore

rm -rf .git

Initialized empty Git repository in /private/tmp/release2/.git/
warning: CRLF will be replaced by LF in vendor/laminas/laminas-diactoros/.laminas-ci.json.
The file will have its original line endings in your working directory
warning: CRLF will be replaced by LF in vendor/laminas/laminas-feed/CHANGELOG.md.
The file will have its original line endings in your working directory
warning: CRLF will be replaced by LF in vendor/laminas/laminas-zendframework-bridge/CHANGELOG.md.
The file will have its original line endings in your working directory
```

I know this is an interaction between the macOS case-insensitive filesystem and settings for handling how to manage files in the working tree. I don't think there's anything to change, but wanted to call it out in case.

I noticed that running this against a built site was taking my laptop around 60s to run the whole task. As compared to rsync using `--files-from`, I'd ballpark around 2 - 2.5x slower. For example, a clean Drupal project is ~9s using rsync with an excludes file, and ~22s with the git method.

Actually, it looks like it could be worse with more / larger files. With a client project that's ~550MB, `snapshot:directory` took 1:26, whereas rsync took 27s, or around 3.2x slower. My guess is much of the perf hit is `git add`... which we'd have to do before deploying for clients on Pantheon or Acquia anyways.

_However_ working and tested by the two of us over the week is better than "possibly faster but not tested". So, my suggestion is we merge this in, and re-evaluate once we have a client who's not using git for deployments.

Issue #11, opened 2021-06-01
Add a process (or just document how) to validate a Taskfile


justafish:
JSON Schema is available https://taskfile.dev/#/community?id=json-schema

PR #17, opened 2021-06-28
Issue #11: Add JSON Schema for Taskfile.yml and an example command to…

justafish:
… validate it


.github/workflows/validate-taskfile.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  Validate-Example-Taskfile:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - run: |
+          curl -O https://json.schemastore.org/taskfile.json
```
deviantintegral, line 18: I'm surprised that this isn't pinned to a version as well.

scaffold/Taskfile.yml
```diff
@@ -4,3 +4,8 @@ dotenv: ['.env']
 
 includes:
   drupal: ./vendor/lullabot/drainpipe/tasks/drupal.yml
+
+tasks:
+  default:
+    cmds:
+      - echo "Hello, world!"
```
deviantintegral, line 11: newline?

I wonder how useful this is, given we could point users to https://taskfile.dev/#/usage. If we keep it, we should add a description so `task -l` doesn't error out.

.github/workflows/validate-taskfile.yml
deviantintegral: Actually, should we just commit this to our repo? `schemastore.org` isn't loading for me, I doubt this would change much, so perhaps save a tiny bit of time?

deviantintegral (approved):
If only we could write a task to validate our task 😂 

This is so cool - hopefully this means that we can easily catch typos like "cmd" vs "cmds" which is still valid yaml.

.github/workflows/validate-taskfile.yml
```diff
@@ -0,0 +1,19 @@
+name: Valiate example Taskfile.yml
```
q0rban, line 1: `s/Valiate/Validate/`

scaffold/Taskfile.yml
justafish: The Taskfile won't validate without at least one command

.github/workflows/validate-taskfile.yml
justafish: I had it committed initially, but I thought this would be easier in the long run. Let's see how stable schemastore.org is and we can check it in if it's consistently wobbly

scaffold/Taskfile.yml
deviantintegral: That's unfortunate, given that there are other tasks included. Worth opening an upstream bug? Task will work without a task defined, so I wonder if it's just a schema issue.

Issue #7, opened 2021-05-19
Select compression and archive format for archives


deviantintegral:
- Broadly available on macOS, Windows, Linux
- Do we optimize for CPU or storage?
- Should support multithreaded compression / decompression
- My memory is that gzip, zip, and bzip2 are all vastly outclassed in both speed and compression by `xz` and other newer formats

PR #18, opened 2021-07-05
Issue #7: Allow xz and bzip2 compression formats to be used for snaps…

justafish:
…hot:archive and update snapshot commands to explicitly require -o=<filename>

### DIFF
--- a/tasks/snapshot.yml
+++ b/tasks/snapshot.yml
@@ -15,39 +15,41 @@ vars:
 
 tasks:
   directory:
-    desc: "Creates a snapshot of the current working directory e.g. task snapshot:directory -- /tmp/release"
+    desc: "Creates a snapshot of the current working directory"
     summary: |
       Creates a snapshot of the current working directory
 
       .git, .gitignore, files listed in .drainpipeignore, and .drainpipeignore
       itself are not added. .drainpipeignore uses the same format as gitignore.
 
-      The target directory should be provided as an argument:
-      task snapshot:directory -- /tmp/release
+        usage: task snapshot:directory o=/tmp/release
+
+        o=<file>   Write the archive to <file>
     cmds:
-      - if [ -d "{{.CLI_ARGS}}" ] || [ "" == "{{.CLI_ARGS}}" ]; then echo "Please provide a path to a directory that does not yet exist" && exit 1; fi
-      - cp -pR . {{.CLI_ARGS}}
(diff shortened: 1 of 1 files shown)


tasks/snapshot.yml
```diff
@@ -15,39 +15,41 @@ vars:
...
       Creates a snapshot of the current working directory
 
       .git, .gitignore, files listed in .drainpipeignore, and .drainpipeignore
       itself are not added. .drainpipeignore uses the same format as gitignore.
 
-      The target directory should be provided as an argument:
-      task snapshot:directory -- /tmp/release
+        usage: task snapshot:directory o=/tmp/release
+
+        o=<file>   Write the archive to <file>
```
deviantintegral, line 27: This is a directory, not a file?

I think renaming `o` could be clearer. How about `output_directory`?

tasks/snapshot.yml
```diff
@@ -15,39 +15,41 @@ vars:
...
       itself are not added. .drainpipeignore uses the same format as gitignore.
 
       If your .gitattributes file contains export-ignore or export-subst, these
       will be respected when exporting the archive.
+
+        usage: task snapshot:archive o=~/archive.tar.bz2
+
+        o=<file>   Write the archive to <file>. The compression format is
+                    inferred from the file extension. Format options are:
+                    tar, tar.bz2, tar.gz, tar.xz, zip
```
deviantintegral, line 52: Missing `.`?

tasks/snapshot.yml
```diff
@@ -58,4 +60,7 @@ tasks:
         git config user.email 'no-reply@lullabot.com'
         git config user.name 'Lullabot Drainpipe'
         git commit -m "Initial commit" --no-gpg-sign
-        git archive --format={{.format}} HEAD
+        git config tar.tar.xz.command "xz -c"
```
deviantintegral, line 63: Interesting we have to set these manually. Is that because the git config is dynamic, and would support us creating a `tar.asdf` file if we defined it?

tasks/snapshot.yml
```diff
@@ -58,4 +60,7 @@ tasks:
...
+        git config tar.tar.bz2.command "bzip2 -c"
+        git archive -o {{.o}} HEAD
+        test -f {{.o}}
```
deviantintegral, line 66: Interesting, does git archive not give a reasonable exit code?

deviantintegral (approved):
A few notes, nothing significant though. Feel free to merge once we've got `o` renamed to something clearer.

tasks/snapshot.yml
justafish: yep!

PR #19, opened 2021-07-05
Integrate lullabot/drainpipe-installer into this repository

justafish:
It seems like we can integrate the composer plugins directly into this repository just fine (unless I'm missing something!)

- Moves the code from https://github.com/Lullabot/drainpipe-installer to here (collapsing the Taskfile installer into one class)
- Updates the gitignore file creation to just check if the text `.task` is in there if `.gitignore` already exists
- Adds a test for the binary installer
- Hooks up the test to GitHub Actions
- Removes the symlink of Taskfile.yml because it's annoying for local testing at the moment, but will follow up on this :smile:


deviantintegral:
I was under the impression it had to be a separate repository so the plugin was installed first and available before other code used it. Otherwise, how would we specify both the type of `composer-plugin` and `drainpipe-installer` in composer.json?

justafish:
@deviantintegral seems to work fine. I've tested it locally with a fresh Drupal install and hooked the tests up to GitHub actions

.editorconfig
```diff
@@ -0,0 +1,19 @@
+; top-most EditorConfig file
```
deviantintegral, line 1: 👏

.github/workflows/composer-plugin.yml
```diff
@@ -0,0 +1,29 @@
...
+    types: [opened, synchronize, reopened]
+
+jobs:
+  Test-Composer-Plugin:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
```
deviantintegral, line 16: Not a blocker for this PR, but we could use 8?

.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
deviantintegral, line 1: 👏 👏

tests/src/Functional/InstallerScaffoldTest.php
```diff
@@ -0,0 +1,48 @@
...
+        parent::setUpBeforeClass();
+        $composer = new Process(['composer', 'update'], self::PROJECT_PATH);
+        $composer->run();
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        parent::tearDownAfterClass();
+
+        //if (file_exists(self::PROJECT_PATH.'/.gitignore')) {
```
deviantintegral, line 25: delete this?

deviantintegral (approved):
Well, this works perfectly! Glad you saw we could collapse into one project.

I saw we are committing composer.lock - that should probably be removed, so tests follow current dependencies?

PR #26, opened 2021-07-09
Hotfix: Check the vendor/bin dir for a binary before trying to overwr…

justafish:
…ite it


src/BinaryInstaller.php
```diff
@@ -207,9 +218,6 @@ protected function installBinary($binary, $version, $url, $sha, $hashalgo = 'sha
         }
 
         if ('.tar.gz' === substr($url, -7)) {
-            // Remove the .tar file from the cache directory if it was
-            // previously extracted.
-            $fs->remove(substr($cacheDestination, 0, -3));
```
rabbitlair, line 212: Is this removed on purpose? It was added in the previous PR, isn't it?

rabbitlair (approved):
Looks good, just a minor question

src/BinaryInstaller.php
justafish: @rabbitlair yes I'm actually already doing it a few lines down already :facepalm:

PR #35, opened 2021-08-17
Don't assume the remote branch already exists when deploying with git

justafish:


tasks/deploy.yml
```diff
@@ -30,4 +34,3 @@ tasks:
         git push origin {{.branch}}
```
justafish, line 36:
```suggestion
```

tasks/deploy.yml
```diff
@@ -20,7 +20,11 @@ tasks:
       - if [ "" == "{{.message}}" ]; then echo "Please provide a commit message" && exit 1; fi
       - |
         TMP_DIR=$(mktemp -d)
-        git clone --depth 1 --branch {{.branch}} {{.remote}} $TMP_DIR
+        git clone --depth 1 --branch {{.branch}} {{.remote}} $TMP_DIR || true
+        if [[ ! -d $TMP_DIR ]]; then
```
justafish, line 24:
```suggestion
        if [[ ! -d "$TMP_DIR/.git" ]]; then
```

rabbitlair (approved):

PR #37, opened 2021-09-21
Allow Symfony 3 for Drupal 8 sites

deviantintegral:
I went to add Drainpipe to The Daily Report, thinking it would be good to do that _before_ we do the Drupal 9 upgrade. However, drainpipe requires `symfony/yaml` 4, and Drupal 8.9 is on Symfony 3. It wouldn't surprise me if we have inherited projects that need to update to Drupal 9 but have some cleanups to do first, even past the official Drupal 8 EOL.

Also, I got to learn about how GitHub actions handles matrix builds, and they're pretty good 👍 

https://github.com/Lullabot/drainpipe/pull/37/checks?check_run_id=3667614911 shows testing against Drupal 8.9, and https://github.com/Lullabot/drainpipe/pull/37/checks?check_run_id=3667615016 tests against 9.2.


justafish:
Great, thanks! :+1:

justafish (approved):

PR #44, opened 2021-10-07
Allow opting out of pushes for local testing

deviantintegral:
When testing locally, I didn't want to pollute the upstream repo. This allows users to not push to the repo by setting `push=false` (or any non-true string).

### DIFF
--- a/tasks/deploy.yml
+++ b/tasks/deploy.yml
@@ -33,6 +33,6 @@ tasks:
         git add -A
         git commit -m "{{.message}}" --allow-empty
         git push origin {{.branch}}
-        if [ "" == "{{.push}}" -o "true" == "{{.push}}" ]; then
+        if [ "{{.push}}" == "" ] || [ "{{.push}}" == "true" ]; then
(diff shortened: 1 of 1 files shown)


tasks/deploy.yml
```diff
@@ -32,4 +33,6 @@ tasks:
         git add -A
         git commit -m "{{.message}}" --allow-empty
         git push origin {{.branch}}
-
+        if [ "" == "{{.push}}" -o "true" == "{{.push}}" ]; then
```
justafish, line 36:
```suggestion
        if [ "{{.push}}" == "" ] || [ "{{.push}}" == "true" ]; then
```
just for consistency with the rest of the codebase - but TIL `-o`!

deviantintegral: I think if you use `||` you need to use `[[`, which has different semantics.

https://linuxhint.com/bash_operator_examples/#o51
https://linuxhint.com/bash_operator_examples/#o17

and https://linuxhint.com/bash_operator_examples/#o17

PR #68, opened 2022-01-26
Hint adding the Drainpipe node packages

hawkeyetwolf:
Could be fancier, such as detecting the missing package to provide a better error, or automatically installing it? But this PR's changes help.


justafish (approved):
Thank you!

PR #73, opened 2022-02-15
Add bitbucket-pipelines.example.yml

hawkeyetwolf:
Just thought I'd drop this here in case we want to bring any Bitbucket Pipelines sample config into Drainpipe. It includes steps to run:
- `test:static`
- `composer-lock-diff` and post annotations back to the PR
- `test:functional`
- Querying Tugboat for Lighthouse reports (thanks to @marcoscano)

The `test:functional` step is not tested and probably isn't working as-is; we're using DTT for functional testing and so I just put that in there as a starter.

The Tugboat Lighthouse step was ported from a separate shell script in our repo... Should work as-is but may be dragons.

I guess to fully flesh out Bitbucket Pipelines support, we'd want to:
- [ ] support a `bitbucket-pipelines` flag in composer "extra" config to copy the scaffolding files just like Drainpipe does for Gitlab CI.

### DIFF
--- a/scaffold/bitbucket/bitbucket-pipelines.example.yml
+++ b/scaffold/bitbucket/bitbucket-pipelines.example.yml
@@ -1,0 +1,157 @@
+image: tugboatqa/php:8.0-apache
+
+pipelines:
+  pull-requests:
+    default:
+      # All steps in the following parallel group will be started at the
+      # same time. These are intended to be quick checks so that if any of them
+      # fails, there is no point in starting the more expensive checks below.
+      - parallel:
+          - step:
+              name: 'Static tests'
+              script:
+                # Prepare the environment.
+                - cp dist/.env.ci .env
+                - composer selfupdate --2
+                - echo "memory_limit = 512M" > $PHP_INI_DIR/conf.d/php-memory-limits.ini
+                # Install dependencies.
+                - composer install --no-ansi --no-interaction --optimize-autoloader --no-progress
+                # Run Drainpipe static tests.
+                - ./vendor/bin/task test:static
+          - step:
+              name: 'composer-lock-diff'
+              caches:
+                - composer
+              script:
+                - apt-get update && apt-get install jq
+                - composer global require davidrjonas/composer-lock-diff:^1.0
+                - COMPOSER_LOCK_DIFF=$(composer global exec -- "composer-lock-diff --from origin/$BITBUCKET_PR_DESTINATION_BRANCH")
+                # Are there any changes to composer.lock?
+                - |
+                  [ -z "$COMPOSER_LOCK_DIFF" ] && exit 0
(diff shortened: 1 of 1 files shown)


scaffold/bitbucket/bitbucket-pipelines.example.yml
```diff
@@ -0,0 +1,156 @@
+image: tugboatqa/php:8.0-apache
```
deviantintegral, line 1: Is this intended to be "copy and never update" or "somehow get updates"? I wonder if it makes sense to rebase on top of https://github.com/Lullabot/drainpipe/pull/123/files#diff-fccaa14362cd8508cb094a15287d2e442fc57d94205c6b2146087a552815d1fcR337 so you can dynamically get the PHP version and set it via twig.

scaffold/bitbucket/bitbucket-pipelines.example.yml
```diff
@@ -0,0 +1,156 @@
...
+      # All steps in the following parallel group will be started at the
+      # same time. These are intended to be quick checks so that if any of them
+      # fails, there is no point in starting the more expensive checks below.
+      - parallel:
+          - step:
+              name: 'Static tests'
+              script:
+                # Prepare the environment.
+                - composer selfupdate --2
+                - echo "memory_limit = 512M" > $PHP_INI_DIR/conf.d/php-memory-limits.ini
```
deviantintegral, line 15: 🤔 I wonder if this should be updated to run ddev inside of pipelines.

scaffold/bitbucket/bitbucket-pipelines.example.yml
```diff
@@ -0,0 +1,156 @@
...
+              name: 'Functional tests'
+              caches:
+                - composer
+              script:
+                - ./vendor/bin/task test:functional
+          - step:
+              name: 'Check Lighthouse reports'
+              script:
+                # Install the Tugboat CLI tool.
+                - wget https://dashboard.tugboat.qa/cli/linux/tugboat.tar.gz
```
deviantintegral, line 110: wut. This is neat! I had no idea we could get the lighthouse data via Tugboat's API.

deviantintegral (commented):
Some thoughts:

1. I do think before merging this should actually be wired into composer as you suggest.
2. I think it could make sense to move the large scripts for lock diffing and lighthouse into Tasks. Then they can be triggered and tested locally. I also think it would make sense to split them somehow into "run the job" and "report the job" components, so only the "report" section would actually require Bitbucket credentials. Then we'd be a step closer to doing things like reporting Tugboat lighthouse results into other VCS tools like GitHub.

deviantintegral:
We don't have any current requests or new projects using Bitbucket. We're closing this PR until that changes. Thanks for sharing it!

PR #78, opened 2022-03-09
Add SASS Glob

ckrina:
I'd love to be able to avoid the extra time defining each component on style.scss, so instead of:

```
// Base
@use "sass/base/base";
@use "sass/base/fonts";
@use "sass/base/forms";
@use "sass/base/tables";
```
We could say

```
// Base
@use "sass/base/**/*";
```

So we avoid extra time and forgetting to add them.


justafish (changes requested):
Great idea!

Could you add something in the README about this?

Looks like you've accidentally checked in your IntelliJ config

ckrina:
@justafish I might have gone too far with documentation, but I think that's what I would need to understand it. But happy to change anything!

README.md
```diff
@@ -56,6 +56,42 @@ See https://github.com/Lullabot/drainpipe-dev
 The best way to specify the browsers to target in the project is a `.browserslistrc` file in the project root. See https://github.com/postcss/autoprefixer
 for more info.
 
+### CSS & JS asset automation
+
+Drainpipe provide default tasks to automate CSS & JavaScript generation with the
+submodules SASS and Javascript.
```
justafish, line 62:
```suggestion
Drainpipe provides tasks to automate Sass & JavaScript compilation.
```

README.md
```diff
@@ -56,6 +56,42 @@ See https://github.com/Lullabot/drainpipe-dev
...
+    - task: javascript:compile
+    - task: sass:compile
+assets:watch:
+  desc: Builds assets such as CSS & JS, and watches them for changes
+  deps: [sass:watch, javascript:watch]
+```
+
+The following commands will be available on the root folder :
+- `ddev task assets` Default task to compile the CSS and JS once for production.
+- `ddev task assets:watch` Task to watch and build the changes while working on the theme.
```
justafish, line 82:
```suggestion
You can then run the following tasks:
- `ddev task assets` Default task to compile the CSS and JS once for production.
- `ddev task assets:watch` Task to watch and build the changes while working on the theme.
```

justafish (approved):

PR #80, opened 2022-03-10
Don't wipe environments when cloning between Pantheon sites

justafish:
Removes the environment wipe and adds `--delete` to the rsync command

### DIFF
--- a/scaffold/github/actions/pantheon/clone-env/action.yml
+++ b/scaffold/github/actions/pantheon/clone-env/action.yml
@@ -27,7 +27,6 @@ runs:
             drainpipe_exec "terminus env:clone-content ${{ inputs.source-site-name }}.${{ inputs.source-environment }} ${{ inputs.target-environment }}"
           else
             TARGET_SITE_ID=$(drainpipe_exec "terminus site:lookup ${{ inputs.target-site-name }}")
-            drainpipe_exec "terminus -y env:wipe ${{ inputs.target-site-name }}.${{ inputs.target-environment }}"
             DATABASE_URL=$(drainpipe_exec "terminus backup:get ${{ inputs.source-site-name }}.${{ inputs.source-environment }} --element=database")
             # Can't pass the URL on the command line directly because it's got weird characters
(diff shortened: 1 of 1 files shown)


alexjain27 (approved):

scaffold/github/actions/pantheon/clone-env/action.yml
```diff
@@ -27,14 +27,13 @@ runs:
...
             # Can't pass the URL on the command line directly because it's got weird characters
             echo $DATABASE_URL > db.env
             drainpipe_exec "terminus -y import:database ${{ inputs.target-site-name }}.${{ inputs.target-environment }} \"\$(cat db.env)\""
             rm db.env
             curl -o files.tar.gz $(drainpipe_exec "terminus backup:get ${{ inputs.source-site-name }}.${{ inputs.source-environment }} --element=files")
             tar -xf files.tar.gz
-            drainpipe_exec "rsync -rLvz --size-only --checksum --ipv4 --progress -e 'ssh -p 2222' ./files_${{ inputs.source-environment }}/. --temp-dir=~/tmp/ ${{ inputs.target-environment }}.$TARGET_SITE_ID@appserver.${{ inputs.target-environment }}.$TARGET_SITE_ID.drush.in:files"
+            drainpipe_exec "rsync -rLvz --delete --size-only --checksum --ipv4 --progress -e 'ssh -p 2222' ./files_${{ inputs.source-environment }}/. --temp-dir=~/tmp/ ${{ inputs.target-environment }}.$TARGET_SITE_ID@appserver.${{ inputs.target-environment }}.$TARGET_SITE_ID.drush.in:files"
```
salvamomo, line 37: I think this is okay :+1: and it _shouldn't_ be an issue for Pantheon envs. We were definitely using it in our blue / green job to get rid of things that could have been added on the inactive environment since the last switch.

Also noticed the `--checksum`. Given it's a clone between two pantheon envs, it may be worth removing it from here since it coud make a difference for large environments.

PR #86, opened 2022-03-15
Force the Javascript target definition

ckrina:
## Problem
While trying to define the JS assets in a new theme I found out that setting the variable `DRAINPIPE_JAVASCRIPT` in `Taskfile.yml` wasn't working properly: whatever I set in there it'd export it in another directory right below `docroot`.

I set:
```
DRAINPIPE_JAVASCRIPT: |
    docroot/themes/custom/my_theme/src/script.js:docroot/themes/custom/my_theme/src/script.min.js
```

And instead of exporting it into `docroot/themes/custom/my_theme/src/script.min.js`, it was creating the files into `docroot`, ignoring the target path.

## Proposed solution
@hawkeyetwolf helped me coming up with this possible solution, forcing the definition of the `outbase` value, and now the source path set is being used for the target too.

But while reading the comments about the [limitation to provide separate entryNames](https://github.com/evanw/esbuild/issues/224), wouldn't it be less confusing to just provide an entry value for any JS on the `Taskfile.yml` and use that to build both for the source and the value path?

@justafish what do you think? Does the changes makes sense? Is it enough, or should we try to improve something else?

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Drainpipe provides tasks to automate Sass & JavaScript compilation.
 
 To enable this, first define the project variables `DRAINPIPE_SASS` and/or
 `DRAINPIPE_JAVASCRIPT` in `Taskfile.yml`.
+You can define several entries, one per line.
 
 Then define the task:
 ```
@@ -90,6 +91,12 @@ It includes SASS Glob to use glob imports to define a whole directory:
 // Base
 @use "sass/base/**/*";
 ```
+
+#### JS
+
+Source and target need to have the same basedir (web or docroot) due to being
+unable to provide separate entryNames.
+See https://github.com/evanw/esbuild/issues/224
 
 ## Validation
 
--- a/metapackages/javascript/esbuild.js
+++ b/metapackages/javascript/esbuild.js
@@ -65,6 +65,7 @@ if (uniqueFileExtension.length !== 1) {
       plugins,
       entryPoints: scripts.map(script => script.split(':')[0]),
       outdir: uniqueBaseDir[0],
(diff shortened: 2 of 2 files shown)


justafish:
@ckrina Would it be possible to setup an example barebones repo reproducing this? We have the following setup on another project and it seems to be working fine:

```
  DRAINPIPE_JAVASCRIPT: |
    web/themes/custom/bookworm/script.js:web/themes/custom/bookworm/script.min.js
    web/themes/custom/ghostwriter/script.js:web/themes/custom/ghostwriter/script.min.js
    web/modules/custom/[client]_design/script.js:web/modules/custom/[client]_design/script.min.js
```

hawkeyetwolf:
@justafish yes to reproducing on a barebones repo. Your example works because of two things:
- the entries use two different directories just below root (`web/themes` and `web/custom`). This causes the `outbase` variable to be set to `web` automatically.
- All the output directories match your source directories.

If you changed either of those things, your JS would compile to unexpected directories. Give a shot at changing either of those and let me know if you see the same. Also take a look at the code and see that the destination directories are not being passed to esbuild.

justafish:
Thanks!

justafish (approved):

hawkeyetwolf:
FYI this still doesn't fix the issue of ignoring the destination directories. Regardless of what destination directory is provided, the compiled JS will always land next to the source file. This is an okay limitation, but we should update the Drainpipe docs to entirely remove the mention of a destination directory. Is that how you'd like to proceed, @justafish, or would you rather we open a new issue for using the destination directories?

justafish:
@hawkeyetwolf let's open a new issue please 👍

PR #101, opened 2022-07-05
Upgrade Terminus to version 3 to support PHP 8

rabbitlair:


justafish:
seems so! https://github.com/pantheon-systems/terminus/issues/2272

justafish (approved):

PR #108, opened 2022-07-18
Support Terminus plugins in Github and Gitlab

rabbitlair:
This PR just makes the [Terminus Secrets plugin](https://github.com/pantheon-systems/terminus-secrets-plugin) available by default when installing Drainpipe

### DIFF
--- a/scaffold/github/actions/pantheon/setup-terminus/action.yml
+++ b/scaffold/github/actions/pantheon/setup-terminus/action.yml
@@ -4,6 +4,9 @@ inputs:
   pantheon-token:
     description: "Pantheon Terminus Token"
     required: true
+  terminus-plugins:
+    description: "Terminus plugins"
+    required: false
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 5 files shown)


scaffold/github/actions/pantheon/setup-terminus/action.yml
```diff
@@ -13,5 +13,6 @@ runs:
         curl -L https://github.com/pantheon-systems/terminus/releases/download/3.0.7/terminus.phar --output ~/terminus/terminus
         chmod +x ~/terminus/terminus
         ln -s ~/terminus/terminus /usr/local/bin/terminus
+        terminus self:plugin:install pantheon-systems/terminus-secrets-plugin
```
deviantintegral, line 16: Should these specify a specific version to install?

deviantintegral (commented):
My only other unrelated thought is that it feels odd to have two different copies of what looks to be the same code for downloading and installing terminus.

deviantintegral:
I was more just thinking that perhaps both CI files call into a shared task or shell script. Regardless, it's not something needed for this PR.

Leaving for @justafish to double check!

justafish (approved):
fantastic! :tada:

Issue #113, opened 2022-08-11
Split the binary downloader out into it's own plugin


deviantintegral:
The binary downloader is great, and it's something I wish I could use on projects by itself. For example, phpmd won't install here without downgrading dependencies, and with a general binary downloader I could grab the phar and stick it into vendor/bin.

deviantintegral:
I'm closing this as won't fix in favour of https://github.com/Lullabot/drainpipe/pull/207.

PR #114, opened 2022-08-11
Refactor installer plugins

deviantintegral:
The plugin code seemed a bit tricky to read with lots of nested conditionals. `phpmd` seems to agree:

> src/BinaryInstaller.php:185          CyclomaticComplexity  The method installBinary() has a Cyclomatic Complexity of 11. The configured cyclomatic complexity threshold is 10.
> src/ScaffoldInstallerPlugin.php:180  CyclomaticComplexity  The method installCICommands() has a Cyclomatic Complexity of 17. The configured cyclomatic complexity threshold is 10.
> src/ScaffoldInstallerPlugin.php:180  NPathComplexity       The method installCICommands() has an NPath complexity of 392. The configured NPath complexity threshold is 200.

With this PR, `phpmd.phar src text codesize` shows no issues other than unrelated PHP deprecation warnings from phpmd itself.

I thought about making a github action for this, but see #113 .


src/BinaryInstaller.php
```diff
@@ -240,4 +240,17 @@ protected function installBinary($binary, $version, $url, $sha, $hashalgo = 'sha
...
+     *
+     * @param string $cacheDestination The destination path to the downloaded file.
+     * @param $hashalgo The hash algorithm used to validate the file.
+     * @param $hash The hash used to validate the file.
+     *
+     * @return bool True if the file needs to be downloaded again, false otherwise.
+     */
+    private function needsDownload(string $cacheDestination, $hashalgo, $hash): bool {
+        return !$this->cache->isEnabled() || !file_exists($cacheDestination) || (file_exists($cacheDestination) && hash_file($hashalgo, $cacheDestination) !== $hash);
+    }
```
deviantintegral, line 255: I think this should have failed as the existing code follows symfony block style. I'll investigate separately.

deviantintegral:
I've pushed up various fixes from IDE checks and phpstan.

I still think this is worth merging, but I wonder if we considered using https://github.com/bamarni/composer-bin-plugin instead? In particular, I was thinking of adding phpstan (which is better as a phar) and the Tugboat CLI, and I wonder if we really do need to maintain our own code.

deviantintegral:
I'm happy to resolve conflicts, but I'd appreciate an initial 👍 on the approach before doing so.

justafish:
We talked about this today and we're going to put it into smaller PRs :tada:

deviantintegral:
I've pulled out a few PRs. Once those are in, I'll do a last one for setting strict types.

mrdavidburns:
@deviantintegral Let's see what happens when this conflict gets resolved. If there's additional work to be done let's create an issue so we can track this task on the project board.

deviantintegral:
The remainder of this is at https://github.com/Lullabot/drainpipe/pull/617.

PR #115, opened 2022-08-22
Runs a config test after doing drupal:update

justafish:
This checks whether or not config has been imported correctly. Particularly useful for PRs opened by Renovate or Dependabot that may require some configuration updates.


justafish:
moved to #141

Issue #96, opened 2022-04-29
Provide a Tugboat template for Pantheon review environments


deviantintegral:
At our Tugboat BoF at the retreat last week, it was brought up that supporting tugboat would help with spinning up new sites and clients. Let's get something in place that sets up a Tugboat configuration file.

deviantintegral:
@rabbitlair @mrdavidburns is there anything in particular you'd like to see in how this works?

deviantintegral:
I have an in-progress PR for the tugboat image up at https://github.com/TugboatQA/images/pull/102. I'd love to pass this off to @justafish to get finished off.

justafish:
Fixed in #213

PR #123, opened 2022-09-18
feat: Add a Tugboat scaffold #96

deviantintegral:
This PR adds support for [Tugboat](https://tugboat.qa/) resolving #96.

To turn it on, add `tugboat` to Drainpipe's `extra` section in `composer.json` following the README such as:

```json
        "drainpipe": {
            "tugboat": {
              "provider": {
                "host": "acquia",
                "downsync": true
              }
            }
        },
```

This PR adds twig as a dependency to assist in rendering configuration files. I first tried to make it so each section was composed "simply" by concatenating files, but that didn't work well. Twig is already part of Drupal so it's not being added completely new. I still refactored the commands out just to save having to edit tugboat command steps directly in YAML.

We talked about pulling in the PHP version from hosting provider files, but that won't cover Acquia hosts. Instead, we use composer to determine the PHP version.

If you set the host to `acquia`, you get mysql and memcached. If you set `pantheon` you get `mariadb` and `redis`.

Testing:

* There are three PRs at https://github.com/Lullabot/drainpipe-tugboat-test/pulls. One has no special setup, one is for Pantheon with Redis, and one is for Acquia with memcache.

I turned on memcache admin and see it working for Acquia. I need to double check that for Pantheon too:

image

* As well, I added some basic tests to the workflow that make sure the right config files are generated. They picked up some good bugs like https://github.com/Lullabot/drainpipe/pull/123/commits/f333b68b5bafbf5cd20380e339b5119366353290. I wonder if later we want to convert these to use `bats` or similar.

To do:

- [x] Automatically add credentials to `settings.php`.
- [x] Add a `tugboat.acquia-example.yml` file that includes memcache.
- [x] Add a `tugboat.pantheon-example.yml` file that includes redis.
- [x] Support setting `acquia` or `tugboat` in the tugboat composer array to select one.
- [x] Configure GD with something like:

```
        apt-get update && apt-get install -y libpng-dev libjpeg-dev libfreetype6-dev
        docker-php-ext-configure gd --with-jpeg --with-freetype && docker-php-ext-install gd
```
- [x] Document in the README what environment variables to set so task can download and import a database.
- [x] Update the configs so if those variables aren't set we run a site:install instead.
- [x] Fix the workflow
- [ ] Switch to nginx for the Pantheon option
- [ ] Ensure we don't have any issues with stale redis caches


deviantintegral:
It turns out Tugboat validates configurations by sending them to a tugboat API. We'll need to generate and provide an access token for the new job to pass.

deviantintegral:
TIL'ed you have to opt in via your job config for repository secrets to be visible. It's not mentioned on the settings UI at all. Eventually I found it at https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow.

edutrul:
The Readme has already been updated, with the new Tugboat variables and it is in a [pull request](https://github.com/Lullabot/drainpipe/pull/132).

And also in the composer, you can support "acquia" and "tugboat"

edutrul:
_Configure GD with something like:
        apt-get update && apt-get install -y libpng-dev libjpeg-dev libfreetype6-dev
        docker-php-ext-configure gd --with-jpeg --with-freetype && docker-php-ext-install gd_

I don't understand what this code would be for, if you can please explain it to me in more detail. Thank you

_Update the configs so if those variables aren't set we run a site:install instead._
When you say site:install, you mean that only the config file will appear in tugboat and the settings

edutrul:
_Update the configs so if those variables aren't set we run a site:install instead._
When you say site:install, you mean that only the config file will appear in tugboat and the settings

deviantintegral:
😮 tests are passing! Well, except the env one which I think is unrelated and being fixed elsewhere.

.github/workflows/test-production-build.yml
```diff
@@ -8,14 +8,17 @@ on:
 
 jobs:
   Test-Production-Build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        php-version: [7.3, 7.4, 8.0]
+        php-version: ["7.3", "7.4", "8.0"]
```
deviantintegral, line 14: Without this actions was treating these as numbers and dropping the `.0` from `8.0`.

.github/workflows/test-production-build.yml
```diff
@@ -40,7 +43,45 @@ jobs:
...
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '18'
+      - name: Install Tugboat CLI for validation
+        run: |
+          cd /usr/local/bin
+          curl -O https://dashboard.tugboatqa.com/cli/linux/tugboat.tar.gz
+          tar xzvf tugboat.tar.gz
+      - name: Validate Tugboat file generation
```
deviantintegral, line 59: These could be turned into matrix parameters in the future, but it's fast now and saves burning CI time on the setup steps.

README.md
```diff
@@ -204,3 +204,54 @@ Requires `GITLAB_ACCESS_TOKEN` variable to be set, which is an access token with
...
+
+### Acquia
+```json
+{
+    "extra": {
+        "drainpipe": {
+            "tugboat": {
+                "provider": {
+                    "host": "acquia",
+                    "downsync": true
```
deviantintegral, line 235: It turns out we don't actually have a task command to downsync from Acquia. @marcoscano or @hawkeyetwolf think that's something you'd like to contribute based on your other project work?

composer.json
```diff
@@ -11,11 +11,13 @@
     },
     "require": {
         "php": "^7.3||^8.0",
+        "ext-json": "*",
```
deviantintegral, line 14: I moved this up as it's common to group php and extensions together.

.github/workflows/test-production-build.yml
justafish: :woman_facepalming:

deviantintegral:
I'm currently working on building a tugboat nginx + php image. I have it building, but nginx is not fully configured yet. Once I have _something_ I can point the tugboat image too I will update this PR.

Issue #124, opened 2022-09-19
Enable .env file support for DDEV


larsdesigns:
Currently ddev does not automatically load the .env file variables. This functionality was removed a few releases back. 

We can enable ddev to load environment variables using a .env file using this the docker-compose ddev extensions technique as documented here: https://ddev.readthedocs.io/en/latest/users/extend/customization-extendibility/#providing-custom-environment-variables-to-a-container using drainpipe. 

This would potentially help us load environment variables for use in testing frameworks that run in Github Actions.

larsdesigns:
Opened pull request: https://github.com/Lullabot/drainpipe/pull/125. The pull request needs testing but I think it will work consistently.

larsdesigns:
No further tasks. Closing...

PR #125, opened 2022-09-19
Enable .env File Support

larsdesigns:
#124 

This pull request implements .env file support using ddev's docker-compose web environment as [documented](https://ddev.readthedocs.io/en/latest/users/extend/customization-extendibility/#providing-custom-environment-variables-to-a-container). 

The docker-compose file is part of the ddev scaffold and is copied into ddev when the ddev commands are also copied into ddev. 

A default .env file is created when a .env file is not found because a .env file is required when enabling the support for the .env file. An error occurs when the .env file is not found. 

### Test Instructions

1. This is testable when drainpipe is installed when ddev project is created or started with drainpipe enabled. 
2. Create a .env file in the root of a ddev project. 
3. Start ddev and use the command `ddev exec 'env' | grep 'your_variable`. 

This PR is a proof of concept. Suggestions and ideas are welcomed and encouraged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ cd drupal
 ddev config
 ddev start
 ddev composer require lullabot/drainpipe
+ddev composer config extra.drupal-scaffold.gitignore true
+ddev composer config --json extra.drupal-scaffold.allowed-packages \[\"lullabot/drainpipe-dev\"]
 ddev composer require lullabot/drainpipe-dev --dev
 # Restart is required to enable the provided Selenium containers
 ddev restart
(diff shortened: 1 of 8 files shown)


mrdavidburns:
I can confirm that this approach was used on Evergreen repo where we are now using Drainpipe, and the environment variables are working. Using `.env` seems like a common thing we have done for most projects in the past and something I'd like for us to continue using. It would be nice if this was baked into Drainpipe so that it's available to all projects using it.

justafish (approved):

justafish:
Thanks!

PR #130, opened 2022-10-17
DRAFT: 96 tugboat acquia

edutrul:
Add a `tugboat.acquia-example.yml` file that includes memcache.

- [ ] Go to composer.json and place:
```
"extra": {
     "drainpipe": {
         "tugboat": ["acquia"]
     }
}
```

- [ ]  run the command

What it does is it creates a file `tugboat.acquia-example.yml`, which contains the memcached.

### DIFF
--- a/scaffold/tugboat/tugboat.acquia-example.yml
+++ b/scaffold/tugboat/tugboat.acquia-example.yml
@@ -1,0 +1,106 @@
+# This serves as a Tugboat configuration file for Drupal websites. It sets up:
+#
+# 1. PHP 8.1 with Apache as the web server.
+# 2. MariaDB 10 for the database.
+# 3. drush-launcher
+# 4. Enables the PHP opcache and required Apache extensions.
+services:
+
+  # What to call the service hosting the site.
+  php:
+    # Only allow access via https.
+    http: false
+
+    # Use PHP 8.1 with Apache
+    image: tugboatqa/php:8.1-apache-bullseye
+
+    # Set this as the default service. This does a few things
+    #   1. Clones the git repository into the service container
+    #   2. Exposes port 80 to the Tugboat HTTP proxy
+    #   3. Routes requests to the preview URL to this service
+    default: true
+
+    # Wait until the mariadb service is done building
+    depends:
+      - mariadb
+
+    # A set of commands to run while building this service
+    commands:
+
+      # Commands that set up the basic preview infrastructure
+      init: |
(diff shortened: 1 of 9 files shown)


scaffold/tugboat/tugboat.acquia-example.yml
```diff
@@ -0,0 +1,106 @@
...
+      # Commands that import files, databases,  or other assets. When an
+      # existing preview is refreshed, the build workflow starts here,
+      # skipping the init step, because the results of that step will
+      # already be present.
+      update:
+        - echo "Nothing to do as we import databases from the Drupal service."
+
+  memcached:
+
+    image: tugboatqa/memcached:1
```
deviantintegral, line 106: Wow, we literally only need two lines here? I figured there would be a tiny bit more.

Seeing this, part of me thinks we could instead combine multiple files together, so we don't have to duplicate the first 100 lines. However, I'm not sure that we won't need _something_ custom for environment variables to do something like downsync Acquia databases into base previews.

src/ScaffoldInstallerPlugin.php
```diff
@@ -273,6 +273,15 @@ private function installCICommands(): void
                     }
                 }
             }
+            foreach ($this->extra['drainpipe']['tugboat'] as $tugboat) {
+                if ($tugboat === 'acquia') {
```
deviantintegral, line 277: What do you think about renaming `$tugboat` to `$provider`?

edutrul: Yeah you are right at the end they are providers and we can have as many other providers. Perfect makes sense thank you, I will do it.

scaffold/tugboat/tugboat.acquia-example.yml
edutrul: _Seeing this, part of me thinks we could instead combine multiple files together, so we don't have to duplicate the first 100 lines. However, I'm not sure that we won't need something custom for environment variables to do something like downsync Acquia databases into base previews._

I don't understand your comment, could you please explain a little more. Thank you

q0rban:
This might be helpful: https://github.com/TugboatQA/docs/issues/207

I would recommend adding the parts to increase the innodb log size and max allowed packet.

edutrul:
Thank you @q0rban I'll add it when no parameter is added like tugboat or acquia. Awesome plan. Thanks again.

PR #131, opened 2022-10-18
DRAFT: Add tugboat pantheon with redis

edutrul:
Add a `tugboat.pantheon-example.yml` file that includes redis.

- [ ]  Go to composer.json and place:
```
"extra": {
    "drainpipe": {
        "tugboat": ["pantheon"]
    }
},
```

- [ ] run the command

What it does is it creates a file `tugboat.pantheon-example.yml`, which contains the redis.


edutrul:
@deviantintegral, I would like to know if you have any ideas why it is failing some checks. Gracias

PR #144, opened 2022-11-02
Add support for Composer Lock Diff to GitHub Actions

justafish:


mrdavidburns:
@justafish How is this progressing? I see it has posted the diff here in the pull request description.
This is a feature that we're really looking forward to for many LSM projects. Is there anything we can do to assist?

PR #147, opened 2022-11-02
Unignore services files that start with wildcard

hawkeyetwolf:
Drupal comes with `development.services.yml`, which is a different filename pattern than what's used for settings files (e.g., `settings.local.php`). 

Optionally, we could unignore _both_ patterns if folks prefer that.

### DIFF
--- a/scaffold/gitignore/common.gitignore
+++ b/scaffold/gitignore/common.gitignore
@@ -27,10 +27,12 @@ web/sites/*
 !web/sites/default
 web/sites/default/*
 !web/sites/default/settings.php
+!web/sites/default/settings.*.php
 !web/sites/default/services.yml
 !web/sites/default/*.services.yml
-!web/sites/default/settings.*.php
 !web/sites/default/files
+# Reginore default services file provided with Drupal scaffolding.
+web/sites/default/default.services.yml
 
 vendor
 
--- a/scaffold/gitignore/common.gitignore
+++ b/scaffold/gitignore/common.gitignore
@@ -28,7 +28,7 @@ web/sites/*
 web/sites/default/*
 !web/sites/default/settings.php
 !web/sites/default/services.yml
-!web/sites/default/services.*.yml
+!web/sites/default/*.services.yml
 !web/sites/default/settings.*.php
 !web/sites/default/files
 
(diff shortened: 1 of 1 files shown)


scaffold/gitignore/common.gitignore
```diff
@@ -28,7 +28,7 @@ web/sites/*
 web/sites/default/*
 !web/sites/default/settings.php
 !web/sites/default/services.yml
-!web/sites/default/services.*.yml
+!web/sites/default/*.services.yml
```
deviantintegral, line 31: The PR title says "ignore"... but this has a negation of `!`?

hawkeyetwolf: Derp 😆 yes, thank you! Updated PR title

deviantintegral (commented):
🤔 so I wonder with this if we want a specific rule to ignore `local.services.yml` or similar. Either way, I'll leave it to @justafish who's thought about this deeper than I have.

hawkeyetwolf:
Yeah, when I used this gitignore recently on a project, I had to add an ignore line for `settings.local.php`... Would be good to have that in there too.

justafish (changes requested):
Thanks for the PR @hawkeyetwolf ! I'd still want to ignore `default.services.yml` as that's controlled by the Drupal scaffolding, anything else would be fine to unignore though

hawkeyetwolf:
> Thanks for the PR @hawkeyetwolf ! I'd still want to ignore `default.services.yml` as that's controlled by the Drupal scaffolding, anything else would be fine to unignore though

Good call. Done! 

Btw what's with unignoring the files directory? I took that line out on my current client project

scaffold/gitignore/common.gitignore
```diff
@@ -26,11 +26,13 @@ web/sites/*
 
 !web/sites/default
 web/sites/default/*
+!web/sites/default/files
```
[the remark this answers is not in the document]
hawkeyetwolf: Yes exactly, do you know why we would want that? This is just moving the existing line, not adding any new/different functionality. I took this line out from a recent client project. My thought is that we can take it out here too. LMK and I'll make the change!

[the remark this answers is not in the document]
justafish: No idea... :laughing: Maybe because of the `.htaccess` file? Let's remove it

justafish, line 29:
```suggestion
```

justafish (approved):

PR #150, opened 2022-11-10
Echo string instead of executing string

beto-aveiga:
### Description

I was having this issue:
image

`Multidev environment hero found, no action taken.: command not found`

Where "hero" was the name of the environment.


justafish (approved):
Thank you!

PR #153, opened 2022-11-17
Remove .env support from DDEV

justafish:
This is causing too many chicken-eggs issues for existing projects - you can't start DDEV without an .env file but the .env file doesn't get scaffolded until you've run `composer install` in DDEV. Given we've got support in task and Drupal, I think this is adequate for now and we can revisit later if it comes up again.

note: turns out the Drupal integration wasn't working properly so this fixes that as well :laughing:


deviantintegral:
Do you have recommendations for projects implementing this themselves so they don’t run into the same issue?

PR #155, opened 2022-11-30
Upgrade the 'cache' github action to v3 to clear warnings about nodejs version

hawkeyetwolf:
I didn't notice any regressions on a client project after upgrading the `actions/cache` action from v2 to v3. It cleared 4 warnings that were present in the UI.


justafish (approved):
Thanks!

PR #163, opened 2022-12-08
Fix a small typo in GitHub action name

hawkeyetwolf:


deviantintegral (approved):
I can't wait until I can use third party epp stores on my iPhone.

PR #166, opened 2022-12-15
Re-use test:config task in drupal:update

hawkeyetwolf:
Just something I noticed. Not sure if there was a reason for not using `test:config` in the `drupal:update` task? Seemed to work like this for me locally.

### DIFF
--- a/tasks/drupal.yml
+++ b/tasks/drupal.yml
@@ -65,7 +65,7 @@ tasks:
       - ./vendor/bin/drush {{.site}} --yes config:import
       - ./vendor/bin/drush {{.site}} --yes updatedb --no-cache-clear
       - ./vendor/bin/drush {{.site}} --yes cache:rebuild
-      - task: :test:config
+      - task: test:config
   maintenance:on:
     desc: Turn on Maintenance Mode
     cmds:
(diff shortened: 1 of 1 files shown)


tasks/drupal.yml
```diff
@@ -65,12 +65,7 @@ tasks:
       - ./vendor/bin/drush {{.site}} --yes config:import
       - ./vendor/bin/drush {{.site}} --yes updatedb --no-cache-clear
       - ./vendor/bin/drush {{.site}} --yes cache:rebuild
-      - |
-        if [[ $(./vendor/bin/drush config:status --format=json --state=Different) != '[]' ]]; then
-          echo "Config export does not match database."
-          ./vendor/bin/drush config:status
-          exit 1;
-        fi
+      - task: :test:config
```
deviantintegral, line 68: Fixing this typo and then hopefully things pass!

```suggestion
      - task: test:config
```

deviantintegral:
Still failing. Conceptually seems fine, so back to you @hawkeyetwolf .

hawkeyetwolf:
@deviantintegral fyi the leading `:` on the task name is for calling Tasks that don't live in the current file and aren't directly included (i.e., they belong to the overall project). 

I knew that was the case here but it didn't ring a bell that it introduces a dependency on `drainpipe-dev`, so closing this... 😆

PR #168, opened 2022-12-19
Support quotes in the Pantheon deploy commit message

hawkeyetwolf:
Commit messages with quotes in them were breaking due to the GitHub Actions template string replacement of the input variable. The quotes in the input variable were interacting with the quotes in the command string.

I haven't tested these actions directly as we're using our own deploy action (modified from Drainpipe's) but it's the same code :)


hawkeyetwolf:
ehh, this didn't seem to fix it in all cases, so closing this PR

PR #179, opened 2023-03-02
Fix variable that provides commit messages in GitHub Actions

justafish:
This value isn't always populated, which causes the deployment to fail due to an empty commit message


deviantintegral (approved):
Makes sense. Not sure how you want testing done?

justafish:
finally! :smile:

PR #184, opened 2023-03-08
Issue #000: Adding Site Schema test.

alexis-saransig-lullabot:
Adds Site Schema testing.
Link: https://github.com/eiriksm/drush-site-schema

### DIFF
--- a/scaffold/Taskfile.yml
+++ b/scaffold/Taskfile.yml
@@ -27,6 +27,19 @@ tasks:
     deps: [drupal:composer:production]
     cmds:
       - echo "Nothing to do"
+  site-schema:
+    desc: Updates site-schema.yml with the latest schema in DB (update and post_update hooks).
+    cmds:
+      - ./vendor/bin/drush {{.site}} site-schema --format=yaml > site-schema.yml
+  drupal:export:
+    desc: The opposite of drupal:update, exports the site's configuration and schema to code.
+    cmds:
+      - composer update --lock
+      - ./vendor/bin/drush {{.site}} --yes cache:clear plugin
+      - ./vendor/bin/drush {{.site}} --yes updatedb --no-cache-clear
+      - ./vendor/bin/drush {{.site}} --yes cache:rebuild
+      - ./vendor/bin/drush {{.site}} --yes config:export
+      - task: site-schema
 # assets:
 #   desc: "Builds assets such as CSS & JS"
 #   cmds:
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
         "drush/drush": "^10|^11",
         "symfony/yaml": "^3|^4|^5|^6",
         "vlucas/phpdotenv": "^4|^5",
-        "ext-json": "*"
+        "ext-json": "*",
+        "eiriksm/site-schema": "^1.1"
     },
     "require-dev": {
         "composer/composer": "^2.0",


scaffold/Taskfile.yml
```diff
@@ -27,6 +27,19 @@ tasks:
...
+      - ./vendor/bin/drush {{.site}} site-schema --format=yaml > site-schema.yml
+  drupal:export:
+    desc: The opposite of drupal:update, exports the site's configuration and schema to code.
+    cmds:
+      - composer update --lock
+      - ./vendor/bin/drush {{.site}} --yes cache:clear plugin
+      - ./vendor/bin/drush {{.site}} --yes updatedb --no-cache-clear
+      - ./vendor/bin/drush {{.site}} --yes cache:rebuild
+      - ./vendor/bin/drush {{.site}} --yes config:export
+      - task: site-schema
```
justafish, line 42: I'm a bit confused about how this task would be used as it seems to run some of the update process - do you have an example workflow that it would be used in please?

justafish (changes requested):

scaffold/Taskfile.yml
mrdavidburns: Here's the documentation from another project where this has been integrated.

```
Site-schema check
This check is based on https://github.com/eiriksm/drush-site-schema , and is intended to verify if the site-schema.yml file is up-to-date with the information from the database. When there is a mismatch, this can be an indicator that DB updates weren't executed or their changes might not be accurately captured in version control.

In order to fix failures from this check, devs are expected to make sure the command:

lando drush @[SITE].local site-schema --format=yaml > site-schema.yml
is executed whenever there is a code change that includes new DB update hooks.

This means executing the command and committing the changes:

After installing a new module (core, contrib, or custom)
After [updating Drupal code](https://bitbucket.org/georgiagov/dsga_d8/src/ef2b4456d1dfaca348c06dd029ee68fc7865affb/resources/docs/80-composer.md) (core or contrib)
Note: This check needs a bootstrappable site to check against, and in order to avoid having to build the full stack only for this check, this is executed as part of the "Functional phpunit tests" job in CircleCI, after the environment is ready, and before running the PHPUnit tests.
```

mrdavidburns: @alexis-saransig-lullabot the command you have listed here just exports the site-schema, it's not doing the schema check.

Here's the code we have running on the other project where this is being used.

```
      - run:
          name: Check that the site-schema file is up-to-date
          command: '/var/www/html/scripts/check-site-schema.sh @test.ci'
```

Here is the contents of that file
```
#!/bin/bash

cd /var/www/html || exit

if [[ -z "$1" ]]; then
  echo "This script needs a site alias to run.";
  exit 42;
else
  SITE_ALIAS=$1;
fi

# We need an already-committed schema to begin with.
if [ ! -f "site-schema.yml" ]; then
    echo "The site-schema.yml file does not exist. Aborting."
    exit 42;
fi

# Export the current schema from DB into a new file.
./vendor/bin/drush "$SITE_ALIAS" site-schema --format=yaml > site-schema-db.yml
# Compare the two.
echo "Comparing site-schema between DB and exported file."
diff -C 5 site-schema.yml site-schema-db.yml > diff.txt
if [ -s diff.txt ]; then
  echo "There are differences between the schema in DB and in the site-schema file. Did you update code without exporting config and/or updating the schema file?"
  echo "Differences:"
  cat diff.txt
  exit 42
else
  echo "No differences found between DB and the site-schema file."
fi
```

justafish:
We discussed this today and opened this as a follow up: #529 - so closing this PR for now

Issue #190, opened 2023-03-31
Run cron once after Drupal update steps


deviantintegral:
Some Drupal modules, like xmlsitemap, install a cron hook to build them. It's common to want to generate those once after pulling down a database or deploying to production.

Sites can do this with their local taskfile with something like this, but I think it's a common enough use case that it's worth doing by default.

```yaml
  override:drupal:update:
    desc: Always run composer before updating.
    deps: [drupal:update]
    cmds:
      # Run cron after updating to refresh items like the XML Sitemap.
      - ./vendor/bin/drush {{ .site }} cron
````

marcoscano:
Could this be opt-in? I see the value when doing this after importing a DB, but it could become an annoyance to devs that run the `drupal:update` task frequently, for example when changing branches, without fully re-importing the DB.

elvism-lullabot:
Created PR #196, when you have a moment please check

mrdavidburns:
For LSM when reviewing site health we check the status reports page. Most of the time without this we're seeing Cron not being run recent enough. Running it with the build steps ensures it has ran recently AND allows us to identify if there are any issues with Cron. Usually these things are not caught until they hit production environments. I see a lot of value having this as part of our build steps and should fail if cron fails.

justafish:
Fixed in #196

PR #196, opened 2023-04-05
Issue #190: Run cron once after drupal update steps.

elvism-lullabot:
See #190

### DIFF
--- a/tasks/drupal.yml
+++ b/tasks/drupal.yml
@@ -72,7 +72,7 @@ tasks:
           exit 1;
         fi
       # Run cron after updating to refresh items.
-      - ./vendor/bin/drush {{ .site }} cron
+      - ./vendor/bin/drush {{ .site }} cron --verbose
   maintenance:on:
     desc: Turn on Maintenance Mode
     cmds:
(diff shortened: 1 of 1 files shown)


tasks/drupal.yml
```diff
@@ -71,6 +71,8 @@ tasks:
           ./vendor/bin/drush {{.site}} config:status
           exit 1;
         fi
+      # Run cron after updating to refresh items.
+      - ./vendor/bin/drush {{ .site }} cron
```
mrdavidburns, line 75: I'm in favor of adding `--verbose` to this. I've found when it fails (which is one of the reasons why we're running this here) we don't get a lot of details about what specifically failed without having `--verbose`

elvism-lullabot: Added `--verbose` option

tasks/drupal.yml
```diff
@@ -71,6 +71,8 @@ tasks:
           ./vendor/bin/drush {{.site}} config:status
           exit 1;
         fi
+      # Run cron after updating to refresh items.
+      - ./vendor/bin/drush {{ .site }} cron --verbose
```
justafish, line 75:
```suggestion
      - ./vendor/bin/drush {{ .site }} core:cron --verbose
```

justafish:
Thanks!

justafish (approved):

Issue #208, opened 2023-05-10
drainpipe dev binaries are missing in the drainpipe-dev repository


deviantintegral:
We're unable to upgrade to Drainpipe 3.x due to this error:

```console
$ ddev composer install
Gathering patches for root package.
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Nothing to install, update or remove
Package doctrine/reflection is abandoned, you should avoid using it. Use roave/better-reflection instead.
Package symfony/debug is abandoned, you should avoid using it. Use symfony/error-handler instead.
Generating autoload files
    Skipped installation of bin bin/drainpipe-twig-linter for package lullabot/drainpipe-dev: file not found in package
    Skipped installation of bin bin/drainpipe-convert-to-junit-xml for package lullabot/drainpipe-dev: file not found in package
```

This breaks `test:static` and `test:lint` with `[test:lint] stat /var/www/html/vendor/bin/drainpipe-twig-linter: no such file or directory`.

PR #209, opened 2023-05-12
(fix) Wrong directory for drainpipe-dev binaries

deviantintegral:
Fixing https://github.com/Lullabot/drainpipe/issues/208 in site-specific code was a pain because the binaries need to be wrapped by composer install for the autoloader to work directly. I think I can add this branch to the push list and use that for testing.


deviantintegral:
https://github.com/Lullabot/drainpipe-dev/tree/208--drainpipe-dev-bin was pushed! It worked for me locally.

Aside, the `Test-Env` job is flakey. It's failed twice on PRs and a rebuild has fixed it.

justafish (approved):

PR #216, opened 2023-06-20
Fixes PHPStan configuration

justafish:


scaffold/phpstan.neon
```diff
@@ -1,2 +1,13 @@
 includes:
-	- vendor/lullabot/drainpipe-dev/config/phpstan.neon
+    - ../../../mglaman/phpstan-drupal/extension.neon
+    - ../../../phpstan/phpstan-deprecation-rules/rules.neon
```
deviantintegral, line 3: @justafish what's the reason for changing these? Will the previously shipped phpstan.neon be removed in drainpipe 4? Wondering what your recommendations are for sites that have customized this file.

tasks/test.yml
```diff
@@ -95,11 +95,15 @@ tasks:
...
         {{ .FUNC_ENSURE_DIRS }}
       - |
+        CONFIG="phpstan.neon"
+        if [ ! -f phpstan.neon ]; then
+          CONFIG="vendor/lullabot/drainpipe/scaffold/phpstan.neon"
+        fi
         if [ "{{.format}}" == "junit" ]; then
           mkdir -p test_result
-          ./vendor/bin/phpstan analyse --error-format=junit {{.TEST_DIRS}} > test_result/phpstan.xml
+          ./vendor/bin/phpstan analyse --error-format=junit -c $CONFIG {{.TEST_DIRS}} > test_result/phpstan.xml
```
deviantintegral, line 104: @justafish This change broke a site that had a `phpstan.neon.dist` file. Since `-c` is being used, this skips the default behaviour documented at https://phpstan.org/config-reference:

> If you do not provide a config file explicitly, PHPStan will look for files named phpstan.neon, phpstan.neon.dist, or phpstan.dist.neon in the current directory.

Two thoughts:

1. Could we instead copy to phpstan.neon.dist? That would reserve `phpstan.neon` for locals and other customized environments, matching phpunit behaviour.
2. If phpstan.neon(.dist) has been excluded form the scaffold via composer.json, then should we just abort the test with a warning and say "phpstan isn't configured"? I'm trying to figure out why the default case here would be triggered.

PR #218, opened 2023-06-21
Force to Use strict types in PHP code

plopesc:
https://architecture.lullabot.com/adr/20211201-use-strict-types-in-php-code/


justafish (approved):
Thanks!

deviantintegral:
How will this apply to existing code we inherit that doesn't have strict types enabled? Is there a way to get phpstan to only evaluate new or changed files for a tests?

justafish:
I guess no differently from code we inherit that violates any other rule - fix it when bringing in this new rule, or disable the test/add exceptions in comments until it can be fixed

justafish:
Test isn't passing due to #226

PR #219, opened 2023-06-23
Ignore new static() PHPStan errors

plopesc:
Ignore these errors, as suggested in core's `phpstan.neon.dist` file

### DIFF
--- a/scaffold/phpstan.neon
+++ b/scaffold/phpstan.neon
@@ -3,6 +3,9 @@ includes:
     - ../../../phpstan/phpstan-deprecation-rules/rules.neon
 
 parameters:
+    ignoreErrors:
+        # new static() is a best practice in Drupal, so we cannot fix that.
+        - "#^Unsafe usage of new static#"
     excludePaths:
         - *sites/simpletest/*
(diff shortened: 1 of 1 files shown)


justafish:
@plopesc this is breaking static tests if there's no existing instances of it. Do you have an example of this in custom code that isn't solved by https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static ?

Relevant issue: https://www.drupal.org/node/3178534

justafish (changes requested):
@plopesc this is breaking static tests if there's no existing instances of it. Do you have an example of this in custom code that isn't solved by https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static ?

Relevant issue: https://www.drupal.org/node/3178534

plopesc:
@justafish Nope, we can handle them following the suggestions https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static

Trying to make it more generic, but if this is throwing exceptions when there are no ignored errors, this is not a good idea.

Thank you for your review!

PR #222, opened 2023-07-05
Fix Sass compilation and add a test

justafish:
Fixes Sass compilation for npm, yarn classic, and yarn modern - as well as adding a test for all 3


justafish:
\o/

PR #227, opened 2023-07-17
Set an .env file from .env.defaults

beto-aveiga:


justafish:
Thanks for the PR - by default however ddev doesn't require a .env to start, this will be something you have set in your specific project. At one point Drainpipe did have this file being read by DDEV, but this was removed due to confusing behaviour and replaced with [this](https://github.com/Lullabot/drainpipe#env-support)

PR #242, opened 2023-08-11
Fix incosistency in deploy.yml file

plopesc:
Deploy task summary says that it accepts a "push parameter to push or not the code to the remote repository.

However, it pushes the code even if this flag is not set to true.

This PR removes the extra push done before the flag checkin.
Also changes the if clause because https://github.com/go-task/task/issues/609 is already fixed.
Finally the command usage example is modified to reflect the actual task behavior


justafish:
@plopesc I didn't mean to merge this just yet, but it caused an issue with the git push for the drainpipe-dev job so I've reverted it here #243

PR #250, opened 2023-08-23
Avoid creation of unwanted node_modules folder as part of test:autofix

plopesc:
This is a followup from #223 

When running `task test:autofix`, the unwanted `node_modules` folder is created and new execution of other tasks that checks the existence of this folder, like `task sass:compile` fail.

This can be mitigated removing manually the `node_modules` before executing the command, but this solution, avoids the folder generation


justafish:
Thanks!

PR #255, opened 2023-08-30
Upgrade Selenium images and replace instaclick/php-webdriver

justafish:
Fixes https://github.com/Lullabot/drainpipe/issues/265


deviantintegral:
From what I can tell, this isn't affecting any clients, but does relate to existing code we support. I'm marking as low for now, unless we get 2+ client sites that are asking for this upgrade.

davereid:
For reference: https://github.com/Lullabot/drainpipe/issues/265

justafish:
This is now in D11, being backported to D10 https://www.drupal.org/project/drupal/issues/3421202

PR #261, opened 2023-09-13
Remove deprecated call to setup_ nodejs script

deviantintegral:
nodesource added a 60 second delay when using these scripts.

image


deviantintegral:
> The maximum number of Multidev environment for this service level has been reached. 

Looks like we need to merge some PRs!

justafish:
@deviantintegral if you run `node --version` in Tugboat now you'll see that it's installed v18 despite us requesting v16

deviantintegral:
@justafish I think that's because we're pulling the node version from ddev, and your test commit was only changing the file used to validate tests. I added a test case for this, does that look good?

justafish:
@deviantintegral the added test is just checking what we've requested - not what is actually installed.

https://github.com/Lullabot/drainpipe/tree/justafish/demonstrate-node-version-problem
https://github.com/Lullabot/drainpipe/blob/justafish/demonstrate-node-version-problem/.tugboat/steps/1-init.sh#L49
https://dashboard.tugboatqa.com/65240cc40bffb36b53917acb

Screenshot from 2023-10-09 15-25-58

deviantintegral:
I think I have this fixed.

On this PR's environment we can see we're now getting node 18 from nodesource instead of node 18 from Debian:

```
root@php:/etc/apt/sources.list.d# apt show nodejs
Package: nodejs
Version: 18.18.0-1nodesource1
Priority: optional
Section: web
Maintainer: NodeSource Support <support@nodesource.com>
Installed-Size: 189 MB
Provides: nodejs-dev, nodejs-doc, nodejs-legacy, npm
Depends: libc6 (>= 2.28), python3, ca-certificates
Conflicts: nodejs-dev, nodejs-doc, nodejs-legacy, npm
Replaces: nodejs-dev (<= 0.8.22), nodejs-legacy, npm (<= 1.2.14)
Homepage: https://nodejs.org
Download-Size: 29.5 MB
APT-Manual-Installed: yes
APT-Sources: https://deb.nodesource.com/node_18.x nodistro/main amd64 Packages
Description: Node.js event-based server-side javascript engine
 Node.js is similar in design to and influenced by systems like
 Ruby's Event Machine or Python's Twisted.
 .
 It takes the event model a bit further - it presents the event
 loop as a language construct instead of as a library.
 .
 Node.js is bundled with several useful libraries to handle server tasks :
 System, Events, Standard I/O, Modules, Timers, Child Processes, POSIX,
 HTTP, Multipart Parsing, TCP, DNS, Assert, Path, URL, Query Strings.

N: There are 24 additional records. Please use the '-a' switch to see them.
```

@justafish can you update your testing branch to confirm the fix is good? I did add a test to the init steps to error out if the installed version doesn't match what's expected. That seemed better than trying to write a test around spinning up a tugboat preview.

justafish:
Screenshot from 2023-10-24 16-51-49

:pray:

PR #268, opened 2023-09-28
Convert variables to support defaults with override #267

deviantintegral:
This lets callers override test directories, as asked for in #267.

Since this is really task functionality, I only added a test for the YAML directories since that seemed to be the simplest to test.

In doing that I discovered something surprising. The symfony yaml library now also ships a `yaml-lint` command, which takes precedence over the one we were using. Since `j13k/yaml-lint` used symfony under the hood, I just removed the dependency.


deviantintegral:
@justafish this is ready for review - tests passing except whatever is going on with nightwatch.

justafish:
Thanks! :+1: :+1:

PR #316, opened 2023-10-27
Update ComposerLockDiff checkout to v4

deviantintegral:
Renovate doesn't touch this file because it's not in the standard GitHub directory.


justafish (approved):
Thanks!

PR #333, opened 2023-11-02
Revert "Update Yarn to v4"

mrdavidburns:
Reverts Lullabot/drainpipe#288

Sorry, I was on the wrong browser tab when I clicked the merge button.


justafish:
I think we can leave this - the issue with Yarn 4 is when using the metapackages in a project with it, but it seems fine here

PR #348, opened 2023-12-06
Disable strict host key checking for Pantheon

justafish:


scaffold/github/actions/common/ddev/action.yml
```diff
@@ -42,6 +42,13 @@ runs:
         fi
         # SSH config file
         touch .ddev/homeadditions/.ssh/config
+
+        # Disable strict host key checking for Pantheon as ssh-keyscan will not
+        # return a stable response.
```
YesCT, line 47: Thanks for the comment explaining why. <3

PR #351, opened 2023-12-08
Add pre-configuration for Nightwatch VRT

justafish:


deviantintegral:
I'm marking this as "hold" because AFAIK none of our existing clients are using Nightwatch VRT or have expressed a desire for it. If we want to do this, let's discuss prioritization as compared to other open issues marked as "client affected".

PR #363, opened 2024-01-29
Remove invalid flag from docker logs

justafish:


deviantintegral (approved):
👍 Makes sense - clearly came from `ddev logs -s` which specifies the service name.

Issue #299, opened 2023-10-25
Tests are broken with Yarn 4


justafish:
```
Run ddev task assets
[assets] ➤ YN0000: · Yarn 4.0.0
[assets] ➤ YN0000: ┌ Resolution step
[assets] ➤ YN0000: └ Completed
[assets] ➤ YN0000: ┌ Fetch step
[assets] ➤ YN0000: └ Completed
[assets] ➤ YN0000: ┌ Link step
[assets] ➤ YN0000: │ ESM support for PnP uses the experimental loader API and is therefore experimental
[assets] ➤ YN0000: └ Completed
[assets] ➤ YN0000: · Done with warnings in 0s 506ms
[sass:compile] node:internal/modules/cjs/loader:1077
[sass:compile]   const err = new Error(message);
[sass:compile]               ^
[sass:compile] 
[sass:compile] Error: Cannot find module 'gulp'
[sass:compile] Require stack:
[sass:compile] - /mnt/ddev-global-cache/yarn/berry/@lullabot-drainpipe-sass-npm-0.1.1-97dbcf6ed2-10c0.zip/node_modules/@lullabot/drainpipe-sass/gulp.js
[sass:compile] Require stack:
[sass:compile] - /mnt/ddev-global-cache/yarn/berry/@lullabot-drainpipe-sass-npm-0.1.1-97dbcf6ed2-10c0.zip/node_modules/@lullabot/drainpipe-sass/gulp.js
[sass:compile]     at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
[sass:compile]     at require$$0.Module._resolveFilename (/var/www/html/.pnp.cjs:11469:54)
[sass:compile]     at Module._load (node:internal/modules/cjs/loader:922:27)
[sass:compile]     at require$$0.Module._load (/var/www/html/.pnp.cjs:11383:31)
[sass:compile]     at Module.require (node:internal/modules/cjs/loader:1143:19)
[sass:compile]     at require (node:internal/modules/cjs/helpers:121:18)
[sass:compile]     at Object.<anonymous> (/mnt/ddev-global-cache/yarn/berry/@lullabot-drainpipe-sass-npm-0.1.1-97dbcf6ed2-10c0.zip/node_modules/@lullabot/drainpipe-sass/gulp.js:1:14)
[sass:compile]     at Module._compile (node:internal/modules/cjs/loader:1256:14)
[...]

PR #383, opened 2024-02-01
#299 Add Yarn 4 to tests

justafish:


justafish:
weirdly this only fails in DDEV, but the process runs fine directly on my local

deviantintegral:
https://github.com/Lullabot/drainpipe/issues/299

PR #413, opened 2024-02-14
Add elasticsearch support to tugboat.

### LINKED ISSUE #387: Add elasticsearch as a preconfigured service to Tugboat
Redis example: https://github.com/Lullabot/drainpipe/blob/main/src/ScaffoldInstallerPlugin.php#L335

KeyboardCowboy:
Resolves #387


.github/workflows/TestTugboat.yml
```diff
@@ -49,7 +51,7 @@ jobs:
       # preview - they should be the same.
       - name: Test Generated Files
         run: |
-          cmp -b drainpipe/.tugboat/config.yml .tugboat/config.yml
```
KeyboardCowboy, line 52: I was trying to find a way to test my changes by generating a more robust config file and testing it against a more complete example.  The test runner doesn't seem to be able to locate this file, though, so I'm not sure if I'm doing something wrong or if there's a better way to test my work.

.gitignore
```diff
@@ -14,3 +14,4 @@ tests/fixtures/drainpipe-test-project/*
 !.yarn/versions
 .npmrc
 lerna-debug.log
+.idea
```
justafish, line 17:
```suggestion
```

.github/workflows/TestTugboat.yml
justafish: @KeyboardCowboy the way we're currently testing it is by having the Tugboat file generated match the one we're using to create a Tugboat environment - bit of a hack, but unfortunately we can't generate a tugboat configuration on the fly. If you add elastic search to `.tugboat/config.yml` then that should be fine. Is there anything we can add to verify that elasticsearch has started correctly and is usable by Drupal?

justafish (changes requested):

.gitignore
KeyboardCowboy: Should I add my IDE settings to the repo?

KeyboardCowboy:
I adjusted the `.tugboat/config.yml` file with the redis and elasticsearch entires to match what the build action should be adding, but the test is failing the file comparison on line 12, which appears the be the empty line between the mariadb dependency and the commands.  Is there a way to view the file the action is generating to compare the two side-by-side?  Otherwise what's the recommended way to test the output?

deviantintegral:
@KeyboardCowboy that's a great question. From what I can tell, since diff also sets an error exit code if it fails we can use that, and get the advantage of not having to replicate this locally.

I pushed up a commit to this PR changing that, and you can see the test error in the CI logs. I'll let you fix it!

KeyboardCowboy:
> @KeyboardCowboy that's a great question. From what I can tell, since diff also sets an error exit code if it fails we can use that, and get the advantage of not having to replicate this locally.
> 
> I pushed up a commit to this PR changing that, and you can see the test error in the CI logs. I'll let you fix it!

Thanks, Andrew.  It looks like it's working now.  Do I need to do something about the sass/JS compilation error?  I'm not sure that has anything to do with my changes.

justafish (approved):

Issue #408, opened 2024-02-12
Add a CI hook to delete old Pantheon multidev environments


deviantintegral:
AFAIK, we don't manually touch multidev environments built by PRs in this repo. We've seen PRs fail due to hitting Pantheon's environment limit.

What if we added a scheduled action that deleted multidevs older than 1 day? Then even if we hit the limit, it should eventually clean up and get PRs back to passing.

justafish:
Sounds like a good idea - a day is probably a bit too short though. Maybe 10 days?

I do go into the dashboard and manually shut stale environments down if they've been open for a while to open some slots

deviantintegral:
I'm marking this as both high and a bug, since it keeps coming up blocking unrelated PRs.

davereid:
Do we have an action that automatically deletes the multidev environment/deployment when the PR is closed/merged?

mrdavidburns:
I believe this happens every time a MultiDev environment is created or updated.
https://github.com/Lullabot/drainpipe/blob/main/scaffold/github/actions/pantheon/review/action.yml#L52

justafish:
We have 20 environments now :tada:

PR #433, opened 2024-02-19
Issue #410: Install the redis PHP extension in Tugboat

### LINKED ISSUE #410: Install the PHP Redis extension when including Redis in Tugboat
We currently detect if redis is used in ddev and add redis as a service in tugboat. However, we don't actually install the PHP Redis extension, so users need to manually add that. Let's add something like:

```
        pecl -q install redis-5.3.7 # Pin a version? Good question!
        docker-php-ext-enable opcache redis
```

to our Tugboat init steps.

justafish:
Fixes #410

### DIFF
--- a/scaffold/tugboat/steps/1-init.sh.twig
+++ b/scaffold/tugboat/steps/1-init.sh.twig
@@ -49,12 +49,16 @@ docker-php-ext-install gd
 # of GD. Lullabot will likely be publishing an ADR recommending it too.
 apt-get install -y imagemagick
 {% if memory_cache_type == "memcached" %}
-# Install the PHP memcache extension. The Drupal Redis module recommends
-# phpredis so no PHP extension is required.
+# Install the PHP memcache extension.
 
 apt-get install -y zlib1g-dev
 pecl install memcache
 echo 'extension=memcache.so' > /usr/local/etc/php/conf.d/memcache.ini
+{% elseifif memory_cache_type == "memcached" %}
+# Install the PHP redis extension.
+
+pecl install redis
+echo 'extension=redis.so' > /usr/local/etc/php/conf.d/memcache.ini
 {% endif %}
 
 # Install node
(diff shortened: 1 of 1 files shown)


scaffold/tugboat/steps/1-init.sh.twig
```diff
@@ -49,12 +49,16 @@ docker-php-ext-install gd
 # of GD. Lullabot will likely be publishing an ADR recommending it too.
 apt-get install -y imagemagick
 {% if memory_cache_type == "memcached" %}
-# Install the PHP memcache extension. The Drupal Redis module recommends
-# phpredis so no PHP extension is required.
+# Install the PHP memcache extension.
 
 apt-get install -y zlib1g-dev
-pecl install memcache
+yes '' | pecl install -f memcache
```
deviantintegral, line 55: Interesting, I've never needed to do any of this in tugboat or CI - just `pecl install <something>`. What's up here?

scaffold/tugboat/steps/1-init.sh.twig
```diff
@@ -49,12 +49,16 @@ docker-php-ext-install gd
...
 echo 'extension=memcache.so' > /usr/local/etc/php/conf.d/memcache.ini
+{% elseif memory_cache_type == "memcached" %}
```
deviantintegral, line 57:
```suggestion
{% elseif memory_cache_type == "redis" %}
```

deviantintegral:
Test failure is maximum multidev limit: https://github.com/Lullabot/drainpipe/issues/408

scaffold/tugboat/steps/1-init.sh.twig
justafish: it should run fine in non-interactive mode, but won't run without asking questions if you put that directly into the tugboat terminal - bit hard for me to test at the moment with our setup so just left it at what I can test directly

deviantintegral (commented):
Looks good to me!

I actually expected this to fail tests though, given there's no corresponding test update.

deviantintegral (approved):

Issue #437, opened 2024-02-19
Configure renovate to not pin dependencies


deviantintegral:
It looks like Renovate is treating our dependencies like an app and trying to pin them. We should either:

- Set renovate to not automatically pin dependencies.
- Set renovate to always widen dependencies.

This should clean up and reduce effort with all of our various dependency updates.

deviantintegral:
I'm calling this high because solving this should significantly reduce the amount of renovate PRs that need manual attention.

I don't have a preference to either of the above solutions - @justafish @davereid can you suggest ways to solve this so we can update the ticket description with a specific action / outcome?

justafish:
This was somewhat addressed already in #416 but we can add some more packages to it

justafish:
Notes from call today:
- Though we don't merge PRs where the range is already wide, e.g. `^6|^7`, it is useful to see tests running on updates to packages to verify they haven't introduced any bugs
- It would be good to record that we're closing them on purpose rather than it looking like we're ignoring the updates

deviantintegral:
I see https://github.com/Lullabot/drainpipe/pull/443 was merged and I think I see what you're doing there - but @justafish mind describing what led you to that solution? And I think that will close this ticket?

mrdavidburns:
All previous PRs closed. 
Renovate isn't pinning anymore.

PR #450, opened 2024-02-23
Issue #437: Just add a label to packages we shouldn't merge

justafish:


renovate.json
```diff
@@ -15,8 +15,7 @@
         "twig/twig",
         "vlucas/phpdotenv"
       ],
-      "matchUpdateTypes": ["minor", "patch"],
-      "enabled": false
+      "addLabels": ["test only - do not merge"]
```
deviantintegral, line 18: This is great! If we ever have trouble with PRs getting merged by mistake, we could probably create an auto-request-changes action too (as Renovate doesn't have this as an option).

PR #457, opened 2024-02-29
Fix variables in setup-ssh GitHub action

jonpugh:
Looks like the variable syntax in the github ssh action isn't right.


justafish:
Thank you!

PR #458, opened 2024-02-29
Fix setup-ssh action apt install commands.

jonpugh:
I'm using this action with a GitHub workflow with `container: drupalci/php-8.1-apache`.

The setup-ssh action doesn't work.

This fixes it.

- Use `apt-get` command.
- Run `apt-get update` to ensure package lists are up to date.
- Use `-y` on install so it works non-interactively.


davereid:
Filed https://github.com/Lullabot/drainpipe/issues/492 about the failing GitHub Action workflow.

justafish:
Thank you!

Issue #385, opened 2024-02-01
Upgrade actions/upload-artifact and actions/download-artifact actions to v4


justafish:
Renovate opened some PRs for this (https://github.com/Lullabot/drainpipe/pull/432), but the tests are broken

https://github.com/Lullabot/drainpipe/pull/378
https://github.com/Lullabot/drainpipe/pull/377

dsayswhat:
Per discussion with @justafish: @penyaskito to look at merge multiple to avoid errors when uploading artifacts

PR #467, opened 2024-03-06
Issue #385: Update actions/upload-artifact and actions/download-artifact to v4

penyaskito:
v3 actions are using node 16, which is EOL, giving some warnings.

This updates:

- actions/upload-artifact
- actions/download-artifact
- actions/cache
- actions/checkout

The artifacts need to be renamed on upload, as they aren't merged by default anymore. See https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md
I've let different artifacts, if we still want to merge the test results is doable, but not sure if we want that.

### DIFF
--- a/.github/workflows/TestComposerLockDiff.yml
+++ b/.github/workflows/TestComposerLockDiff.yml
@@ -75,7 +75,7 @@ jobs:
 
       - name: Upload artifact
         if: ${{ always() }}
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: processed-descriptions
           path: |
--- a/.github/workflows/TestFunctional.yml
+++ b/.github/workflows/TestFunctional.yml
@@ -62,7 +62,7 @@ jobs:
         run: zip -r /tmp/drainpipe.zip ./
 
       - name: Upload Artifact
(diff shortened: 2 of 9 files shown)


penyaskito:
@justafish In v4, Artifacts are immutable (unless deleted). `merge-multiple: true` is on download, not on upload https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md?rgh-link-date=2024-03-06T18%3A41%3A22Z

We could append something like:

```
  Merge-test-results:
    runs-on: ubuntu-22.04
    needs: [Test-PHPUnit-Functional, Test-PHPUnit-Static, Test-PHPUnit-Functional-DTT]
    steps:
     - name: Merge Artifacts
       uses: actions/upload-artifact/merge@v4
       with:
         name: test_results
         pattern: test_results-*
```

to merge them per workflow, but the other artifacts will still be there (or marked as Expired if we set `retention-days: 1`, but still there).

Take into account that this exists only in the repo workflows for testing, not the scafffold ones, so there's no way we can break anything on projects.

mrdavidburns:
This fixes the issues that are happening in #432

mrdavidburns (approved):

justafish:
Thanks!

Issue #464, opened 2024-03-05
Tugboat: support for `online` step


mrdavidburns:
In addition to the `init`, `update`, and `build` steps for Tugboat there's also an `online` step that runs after the build step completes and the container is committed. This is currently not supported in Drainpipe but we do make use of it on some projects. 

https://docs.tugboatqa.com/setting-up-services/how-to-set-up-services/leverage-service-commands/

geekygnr:
As far as I can tell this issue has been completed and is already part of a stable release

PR #471, opened 2024-03-07
Issue #464: Adding support for Tugboat online step.

mrdavidburns:
I've modeled this off of the other Tugboat steps that Drainpipe supports. 
I'm sure there are other pieces that I need to wire up to make this work. 

Looking for some review and guidance on the pieces I'm missing.


justafish:
@mrdavidburns Thanks! I've made the step optional as unlike build, sync, and update it doesn't really map to a required part of the build process. Let me know if that makes sense, as well as the documentation, feel free to merge if so :+1:

Issue #135, opened 2022-10-26
Errors when running composer install: Unable to add newly converted phar task_linux_amd64.tar


beto-aveiga:
I'm trying to run functional tests.
When I run the "Build Project" step, on the first line that runs `composer install`, it fails.
Do you know what is happening?
Thanks!

---
`composer [install] failed, composer command failed: ComposeCmd failed to run 'COMPOSE_PROJECT_NAME=ddev-abc-exampleorg-website docker-compose -f /home/runner/work/abc-example.org-website/abc-example.org-website/.ddev/.ddev-docker-compose-full.yaml exec -w /var/www/html/drupal -T web composer install', action='[exec -w /var/www/html/drupal -T web composer install]', err='exit status 1', stdout='Gathering patches for root package.`

---
`In BinaryInstaller.php line 214:`
`Unable to add newly converted phar "/var/www/html/.ddev/.drainpipe-composer-cache/files/lullabot/drainpipe/bin/task/3.9.0/task_linux_amd64.tar" to the list of phars, a phar with that name already exists`

![image](https://user-images.githubusercontent.com/3916979/198123474-b70757d1-3ade-4a0f-8296-7e7db044fddb.png)

### Workflow code

````
name: "Functional Tests"

on:
  push:
    branches:
      - master
  pull_request:
    types: [opened, synchronize, reopened]

concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
  cancel-in-progress: true

jobs:
  Functional-Tests:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3

      - uses: actions/cache@v3
        with:
          path: ${{ github.workspace }}/.ddev/.drainpipe-composer-cache
          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
          restore-keys: |
            ${{ runner.os }}-composer-

      - uses: ./.github/actions/drainpipe/set-env

      - name: Install and Start DDEV
        uses: ./.github/actions/drainpipe/ddev
        with:
          git-name: Drainpipe Bot
          git-email: no-reply@example.com
          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
          ssh-known-hosts: ${{ secrets.SSH_KNOWN_HOSTS }}

      - name: Build Project
        run: |
          composer install --no-interaction --quiet
[...]

PR #479, opened 2024-03-14
Add a test for the binary installer and fix error when upgrading task binary

### LINKED ISSUE #164: "failed to open stream: Text file busy" while upgrading Task binary
## How
I got the error while upgrading drainpipe from 1.2.16 to 1.3.0, which included an update to the Task binary (https://github.com/Lullabot/drainpipe/pull/158).

## Why
In `vendor/lullabot/drainpipe/src/BinaryInstaller.php` we reduce the likelihood of seeing the "Text file busy" error by only copying it over when necessary. But when upgrading the Task binary, the file must still be replaced, creating an opportunity for the error to arise.

## Ideas
@deviantintegral mentioned:
> perhaps this circular dependency between task and composer should be broken. Drainpipe could validate the task version, but not actually download it (just as you have to download composer out of band to use it).

Thoughts??

justafish:
Here's the test failing for commit 53f88b0 https://github.com/Lullabot/drainpipe/actions/runs/8286318726/job/22676024562

Fixes #164 and #135

### DIFF
--- a/src/BinaryInstaller.php
+++ b/src/BinaryInstaller.php
@@ -210,6 +210,8 @@ class BinaryInstaller implements PluginInterface, EventSubscriberInterface
         }
 
         if ('.tar.gz' === substr($fileName, -7)) {
+            // Remove .tar
+            $fs->remove(substr($cacheDestination, 0, -3));
             $archive = new \PharData($cacheDestination);
             $archive->decompress();
             $archive = new \PharData(substr($cacheDestination, 0, -3));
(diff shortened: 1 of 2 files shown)


src/BinaryInstaller.php
```diff
@@ -210,6 +210,8 @@ protected function installBinary($binary, $version, $url, $sha, $hashalgo = 'sha
         }
 
         if ('.tar.gz' === substr($fileName, -7)) {
+            // Remove .tar
+            $fs->remove(substr($cacheDestination, 0, -3));
```
mrdavidburns, line 214: 💥  👍🏼

mrdavidburns (approved):

PR #495, opened 2024-03-22
Fixes to drainpipe-dev deployments

### LINKED ISSUE #459: Deploying to drainpipe-dev requires the tag to be created manually
The current workflow is pushing a branch instead of a tag

### LINKED ISSUE #460: Switch from a GitHub Token to a deploy key
these don't expire and aren't tied to a user :tada:

### LINKED ISSUE #472: Add repo-lockdown action to prevent new issues and PRs for drainpipe-dev
I think we should add https://github.com/marketplace/actions/repo-lockdown to the drainpipe-dev code to prevent any issues and PRs from being kept open in https://github.com/Lullabot/drainpipe-dev

davereid:
Fixes #459
Fixes #460 
Fixes #472

* Updates the Drainpipe-Dev workflow to no longer need DDEV and just works with raw Git.
* Uses the https://github.com/webfactory/ssh-agent action to add the drainpipe-dev deploy key using the `DRAINPIPE_DEV_DEPLOY_KEY` secret set in this repository.
* Adds a new workflow within drainpipe-dev itself to automate creating the releases when a tag is pushed.
* Adds the https://github.com/marketplace/actions/repo-lockdown workflow in drainpipe-dev to auto-close any issues or PRs filed.

## Remaining Tasks
* [x] Remove this branch from the push event listing

## Follow-ups
* Could we utilize signed commits with https://github.com/planetscale/ghcommit-action? (maybe leave this as a low-priority follow-up)


.github/workflows/DrainpipeDev.yml
```diff
@@ -1,50 +1,60 @@
...
     branches:
+      - 460-drainpipe-dev-use-deploy-token
       - main
     tags:
       - 'v*'
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.ref_name || github.sha }}
   cancel-in-progress: false
```
davereid, line 13: Note head_ref doesn't exist except on pull_request events. So updating this to the branch is at least better.

I'm also curious what happens when we get two back to back quick merges to main, should we use `cancel-in-progress: true` to stop the first one and let the second one run?

.github/workflows/DrainpipeDev.yml
```diff
@@ -1,50 +1,60 @@
...
+          git reset --mixed origin/${{ github.ref_name }}
+          git config user.name "Lullabot-Drainpipe[bot]"
+          git config user.email "157769597+Lullabot-Drainpipe[bot]@users.noreply.github.com"
+
+      - name: Push branch to drainpipe-dev
+        if: ${{ startsWith(github.ref, 'refs/heads/') }}
+        working-directory: drainpipe-dev
+        run: |
+          git add -A
+          git commit -m "$COMMIT_MESSAGE" --allow-empty
```
davereid, line 49: I wonder if the tag also needs the git commit to be done, to ensure we are pushing the correct code to the tag.

.github/workflows/DrainpipeDev.yml
```diff
@@ -8,43 +8,51 @@ on:
       - 'v*'
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.ref_name || github.sha }}
```
justafish, line 11:
```suggestion
  group: drainpipe-dev
```

davereid: We could just use `${{ github.workflow }}` in this case then, since that's a static string too.
```suggestion
  group: ${{ github.workflow }}
```

justafish:
:tada:

PR #520, opened 2024-04-11
Fix composer install with --no-cache flag

justafish:


src/BinaryInstaller.php
```diff
@@ -196,10 +196,15 @@ protected function installBinary($binary, $version, $url, $sha, $hashalgo = 'sha
...
         $fileName = array_pop($parts);
-        $cacheFolder = $this->cache->getRoot().$binary.\DIRECTORY_SEPARATOR.$version;
-        $cacheDestination = $cacheFolder.\DIRECTORY_SEPARATOR.$fileName;
-        $cacheExtractedBinary = $cacheFolder.\DIRECTORY_SEPARATOR.$binary;
-        $binDestination = $bin.\DIRECTORY_SEPARATOR.$binary;
+        if ($this->cache->isEnabled()) {
+            $cacheFolder = $this->cache->getRoot() . $binary . \DIRECTORY_SEPARATOR . $version;
+        }
+        else {
+            $cacheFolder =  sys_get_temp_dir() . \DIRECTORY_SEPARATOR . $binary . \DIRECTORY_SEPARATOR . $version;
```
davereid, line 203:
```suggestion
            $cacheFolder = sys_get_temp_dir() . \DIRECTORY_SEPARATOR . $binary . \DIRECTORY_SEPARATOR . $version;
```

leonel-lullabot (commented):
I tested this solution on the project where I saw the error, and now I can run `composer install --no-cache` without errors. Thanks @justafish!

.github/workflows/TestBinaryInstaller.yml
```diff
@@ -54,3 +54,12 @@ jobs:
...
           ddev composer install
+
+      - name: Test installing without composer cache
+        run: |
+          ddev composer install --no-cache
+          ddev composer install
+          rm -rf vendor
+          ddev composer install
+          rm -rf vendor
+          ddev composer install --no-cache
```
deviantintegral, line 65: I know this tests by making sure there's no errors, but would it make sense to add `test -x vendor/bin/task` too?

src/BinaryInstaller.php
deviantintegral: I actually wonder if in the `--no-cache` case, we should... not cache? This looks to be caching in the system temp directory, which means we're more behaving like "don't cache in the usual composer dir".

justafish: @deviantintegral it won't retrieve it from the tmp directory on the next install and will do a fresh download, it's just being stashed there before being unzipped

.github/workflows/TestBinaryInstaller.yml
justafish, line 65:
```suggestion
          ddev composer install --no-cache
          test -x vendor/bin/task
          ddev composer install
          test -x vendor/bin/task
          rm -rf vendor
          ddev composer install
          test -x vendor/bin/task
          rm -rf vendor
          ddev composer install --no-cache
          test -x vendor/bin/task
```

PR #532, opened 2024-04-16
Fix docs and typos

deviantintegral:
This is extracted from https://github.com/Lullabot/drainpipe/pull/114/commits. It cleans up some phpdocs, and adds a typehint for a protected function.


davereid (approved):
LGTM as long as all the jobs pass

PR #533, opened 2024-04-16
Pull gitlab / github installer code into new methods

deviantintegral:
This PR pulls is the first two commits from https://github.com/Lullabot/drainpipe/pull/114/commits that:

1. Extracts private methods to reduce the length of the installCiCommands method.
2. Adds an early return to manage some of the intense nesting the code currently has.


justafish:
[nesting intensifies]

justafish (changes requested):
@deviantintegral I fixed the database version error mismatch, which I thought was a result of a DDEV upgrade. I think the test failures are now genuine however as the Tugboat files don't seem to be be installed correctly.

scaffold/pantheon/pantheon.yml
```diff
@@ -2,6 +2,6 @@ api_version: 1
 web_docroot: true
 php_version: 8.0
 database:
-  version: 10.4
+  version: 10.11
```
davereid, line 5: How do we feel about this change going out to all sites? Does this upgrade to several versions work okay on Pantheon deployments?

davereid: I feel like I don't see 10.11 listed on https://docs.pantheon.io/pantheon-yml#specify-a-version-of-mariadb so does this need to be 10.6?

Issue #332, opened 2023-11-02
Replace ComposerLockDiff functionality with a sticky PR comment instead


justafish:
Trying to preserve the original content of a comment is causing too many bugs, and doesn't work well when there are edit conflicts

- [ ] Use a sticky pull request comment instead of a section in the PR body.
- [ ] Configure the action to only run in composer.lock is modified: https://stackoverflow.com/a/63822945

davereid:
Related: https://github.com/Lullabot/drainpipe/issues/317

davereid:
My proposal (with approval already from @justafish) is to replace updating the PR body with a sticky comment using https://github.com/marocchino/sticky-pull-request-comment.

mrdavidburns:
@YesCT and I were just talking about this yesterday. We've had multiple projects where we successfully switched to posting to a comment to avoid this issue. +1 for switching Drainpipe posting to comment.

davereid:
Also would be looking to remove DDEV as a dependency of this task to help speed up the job (would also fix https://github.com/Lullabot/drainpipe/issues/239)

deviantintegral:
I updated this to also add a one-line change to only run the action at all if composer.lock is modified.

davereid:
I think it would still need to run on push because it's possible for a PR to have a composer.lock change initially, but then *not* have one after merging, so it would still need to delete the comment in that case. But running the diff could be conditional on changes in that file.

justafish:
here's an alternative solution that combines it with the security checks - 2 birds with 1 stone!
https://github.com/Lullabot/drainpipe/pull/636

Issue #318, opened 2023-10-28
ComposerLockDiff changes line endings, breaking editing PR descriptions in GitHub


deviantintegral:
When editing pull request descriptions, ComposerLockDiff changes `\r\n` to `\n\n`. This makes it very difficult to edit the PR description in the UI, because you have hidden unprintable characters. Top is before (`pull_request.json`), bottom is after (`processed.json`):

<img width="412" alt="image" src="https://github.com/Lullabot/drainpipe/assets/255023/929d60d4-1578-44ca-b905-7f2988956033">

This would be partially solved by #317 moving the contents to a comment, but should also be fixed because `\n` is not supported: https://stackoverflow.com/a/46751663 (most web applications use `\r\n` for line endings): https://www.rfc-editor.org/rfc/rfc7230#section-3.5

justafish:
Closing in favour of #332

beto-aveiga:
If we are going to do this, I can work on a PR to fix the `\n\n \r\n` issue.
@deviantintegral @justafish

YesCT:
@beto-aveiga I'm not sure, but I think you might want to build on the branch from https://github.com/Lullabot/drainpipe/pull/538 . Unless @davereid is also going to work on the line ending thing in his PR.

davereid:
My fix is to move the composer lock diff to a sticky PR comment instead of updating the PR description. The work for this would be fixed by https://github.com/Lullabot/drainpipe/pull/538

davereid:
But we could put in a fix *now* since that's needing some more detailed time and reviews. Ultimately the fix would be deprecated/removed by the work in #538.

mrdavidburns:
Going to move this into Blocked until we move this task out of DDEV.
Complete #283 first

justafish:
Fixed in #636

PR #538, opened 2024-04-22
Update ComposerLockDiff workflow to run without DDEV and use a sticky pull request comment

davereid:
Fixes https://github.com/Lullabot/drainpipe/issues/332

Improvements:
* Action runs now only take about 30 seconds instead of 2+ minutes, no more DDEV install needed
* It will stop early if there are no changes to composer.lock
* It uses https://github.com/IonBazan/composer-diff-action to help run the diff really quickly instead of having to download PHP and install something
* It uses https://github.com/marketplace/actions/sticky-pull-request-comment to post the results back to the PR instead of needing to post to the issue description.

Testing PR:
https://github.com/Lullabot/drainpipe-test/pull/11

Run compares:
* https://github.com/Lullabot/lullabot.com-d8/actions/workflows/ComposerLockDiff.yml (current)
* https://github.com/Lullabot/drainpipe-test/actions/workflows/ComposerLockDiff.yml (this PR)


scaffold/github/workflows/ComposerLockDiff.yml
```diff
@@ -17,46 +17,33 @@ jobs:
...
-        run: |
-          git reset --soft HEAD^1
-          git reset .
+      - name: Generate composer diff
+        if: ${{ steps.composer-lock-changed.outputs.any_changed == 'true' }}
+        id: composer-diff
+        uses: IonBazan/composer-diff-action@v1
+        with:
+          with-platform: true
+          with-links: true
```
deviantintegral, line 40: Does it make sense to also include the underlying CLI command in ddev via a Dockerfile? That way, developers are one step closer to debugging things when they go wrong.

davereid: I don't really think so, and I would actually advocate for removing composer-lock-diff as a DDEV command. I never found a need for it to run this locally in the same way that it does with the action. I could just always debug the action itself or report an issue upstream with the community action.

It is worth noting this is a change from using https://github.com/davidrjonas/composer-lock-diff to https://github.com/IonBazan/composer-diff. There are some pros and cons here with that change:

| composer-lock-diff | composer-diff |
|-|-|
| Run `composer-lock-diff` | Run `composer diff` (composer plugin) |
| Last release Mar 2022 | Last release Dec 2023 |
| No GitHub action available | Maintained GitHub action available |
| Supports Drupal.org diffs | Does not yet support Drupal.org diffs (see https://github.com/IonBazan/composer-diff/issues/24) |

That said, having just the `composer diff` command available from the Dockerfile *was* useful, but I don't think it needs to be a task at all since it would be available as a native composer command.

deviantintegral:
@beto-aveiga is going to review this for https://github.com/Lullabot/drainpipe/issues/318#issuecomment-2117988058 and if there's overlap file a PR against this branch.

YesCT:
We discussed this in the drainpipe sync today. I'd kinda like to see the performance improvement (to not use ddev) and the functional improvement of posting NOT in the description be separate PRs.

An example from mjff ... https://github.com/Lullabot/mirror-mjff/blob/master/.github/workflows/ComposerLockDiff.yml

mrdavidburns:
@beto-aveiga is going to code review and test this branch out on an existing project to determine time saved.

scaffold/github/workflows/ComposerLockDiff.yml
```diff
@@ -17,46 +17,33 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 2
+          fetch-depth: 0
 
-      - uses: actions/cache@v4
+      - name: Check if composer.lock was changed
+        id: composer-lock-changed
+        uses: tj-actions/changed-files@v44
```
mrdavidburns, line 24: @deviantintegral suggests that we pin to a commit hash so future releases don't unexpectedly break our builds.

scaffold/github/workflows/ComposerLockDiff.yml
```diff
@@ -17,46 +17,33 @@ jobs:
...
-          restore-keys: |
-            ${{ runner.os }}-composer-
+          files: 'composer.lock'
 
-      - uses: ./.github/actions/drainpipe/set-env
-
-      - name: Install and Start DDEV
-        uses: ./.github/actions/drainpipe/ddev
+      - name: Delete sticky pull request comment
+        uses: marocchino/sticky-pull-request-comment@v2
```
mrdavidburns, line 29: @deviantintegral suggests that we pin to a commit hash so future releases don't unexpectedly break our builds.

scaffold/github/workflows/ComposerLockDiff.yml
```diff
@@ -17,46 +17,33 @@ jobs:
...
+          delete: true
 
-      - name: Prepare repository
-        run: |
-          git reset --soft HEAD^1
-          git reset .
+      - name: Generate composer diff
+        if: ${{ steps.composer-lock-changed.outputs.any_changed == 'true' }}
+        id: composer-diff
+        uses: IonBazan/composer-diff-action@v1
```
mrdavidburns, line 37: @deviantintegral suggests that we pin to a commit hash so future releases don't unexpectedly break our builds.

scaffold/github/workflows/ComposerLockDiff.yml
```diff
@@ -17,46 +17,33 @@ jobs:
...
-          ddev task "github:composer-lock-diff pull_request='pull_request.json' json_file='processed.json'"
-          curl -f \
-            -X PATCH \
-            -H "Accept: application/vnd.github+json" \
-            -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
-            https://api.github.com/repos/${{ github.repository }}/pulls/$DRAINPIPE_PR_NUMBER \
-            -d @processed.json
+      - name: Post sticky pull request comment
+        if: ${{ steps.composer-lock-changed.outputs.any_changed == 'true' }}
+        uses: marocchino/sticky-pull-request-comment@v2
```
mrdavidburns, line 44: @deviantintegral suggests that we pin to a commit hash so future releases don't unexpectedly break our builds.

mrdavidburns:
@beto-aveiga Instead of pulling this out of DDEV, we discussed bundling it in with the Static Tests which does use DDEV but we're no longer having to wait for another job to build that container since it's already available.

scaffold/github/workflows/ComposerLockDiff.yml
YesCT: How do we get notified of new releases to update the hash to? Monitor the log files for deprecation notices?

davereid:
I'm not sure that bundling it is a wise idea, if static tests fail, or DDEV fails to spin up, we run into the same problem where PRs won't have their diffs posted, which can still be very helpful information to review on a PR while broken jobs get resolved.

scaffold/github/workflows/ComposerLockDiff.yml
davereid: While I agree that locking it down to commit hashes is better for security, the problem becomes when we need to update the workflows to all the consumers, or any repositories haven't been updated in a while, their actions could potentially break. It's a trade-off that we can balance with only doing it for very specific, trusted actions.

Issue #535, opened 2024-04-18
Add MANUAL ONLY pantheon review apps / multidev deploy github action 


YesCT:
Already merged: https://github.com/Lullabot/drainpipe/pull/670 
Related: https://github.com/Lullabot/drainpipe/issues/553
Related/Split to: https://github.com/Lullabot/drainpipe/issues/826 Change the automated action to be dependent on a label (still automated, like on PR update/commit, but also controllable).

Still todo:

- Manual deploy feature: fix, has bug about namiing and character limit. (If there isn't another issue, we can use this issue for that work.)

Notice: ] Logged in via machine token.
Warning: ] You have no multidev environments.
Error: ---------------------------] Create a Multidev environment [error]  The branch pr-MAIN-123 cannot be made into a cloud development environment according to Pantheon's environment naming conventions. The branch name must be a valid domain name with less then 11 characters and no capitalized characters. To make this branch compatible with cloud development environments try renaming the branch to pr-main-123 instead. 

Failed to execute command `terminus multidev:create ***.live pr-MAIN-1230--test-deploy-action --no-db --no-files --yes`: exit status 1
Pantheon puts an 11 character limit on env names. Your name has been truncated to: pr-MAIN-123
Error: Process completed with exit code 1.

----

Original request:

Situation:

site with very limited github minutes, with automatic updater, does not have enough minutes to deploy every PR automatically.

example site 1: PantheonReviewApps.yml is $18 a month
example site 2: 11 repos all sharing 3000 minutes, estimate adding PantheonReviewApps.yml adds 3732 additional minutes, 2115 minutes over 3000 free, adding cost $170 / month. [If we used 3000 of the free for other things, 3732 PantheonReviewApps.yml would add $298 / month.]

Want:

Only deploy branches to multidevs when QA needs them.

Workaround:

Use the pantheon multidev branch deploy from the pantheon dashboard.
Cons: pantheon deploy would not follow https://architecture.lullabot.com/adr/20230929-drupal-build-steps/ 

Request:

- Add config option for MANUAL ONLY pantheon review apps / multidev deploy github action.
- Document choice of setup in ReadMe.

![Screenshot 2024-04-18 at 10 12 16 AM](https://github.com/Lullabot/drainpipe/assets/1970586/d16c6b30-3cb5-4e44-9ea1-026c117582e7)

justafish:
We discussed this and @YesCT is going to do some more research on how much of an impact this is having

[...]

PR #546, opened 2024-04-24
[In Use: update with release] Issue #535: Change multidev deploy to manual only: tag based

leonel-lullabot:
This branch is in use on customer sites.

Please develop using https://github.com/Lullabot/drainpipe/pull/670 . Sorry for the troubles.

### Relates to:
- #535

### Description:
- Run the Multidev Deploy when the label `pantheon-multidev` is added


YesCT (commented):
I'm not sure how to make this configurable. But this PR is good for demonstrating the alternative.

---

Maybe 

```
"extra": {
    "drainpipe": {
        "github": ["PantheonReviewApps"]
    }
}
```

could be the default and 

```
"extra": {
    "drainpipe": {
        "github": ["PantheonReviewAppsManual"]
    }
}
```

could switch to this manual workflow_dispatch ??

I'm not sure what config could look like to send an option to PantheonReviewApps. 

---

Keeping this as Draft is good to indicate this is not ready for merge.

justafish:
Here's the relevant bit of code that will need modifying https://github.com/Lullabot/drainpipe/blob/main/src/ScaffoldInstallerPlugin.php#L285

YesCT:
Ran into an error with the multidev pr number being a string ("renovate") when trying this out on a customer project. There will probs be some more commits here before this is ready to move out of Draft.

YesCT:
We are running this branch on one customer site. I'd like to get reviews from folks before copying it to 10 other repos.

YesCT:
@leonel-lullabot Is there anything here you wanted to update or notes you wanted to leave for reviewers?

scaffold/github/workflows/PantheonReviewAppsManual.yml
```diff
@@ -0,0 +1,72 @@
...
+# You will also need to add the secrets used below to Dependabot secrets
+# if you want this to successfully run in auto-created PRs.
+permissions:
+  deployments: write
+  contents: read
+  pull-requests: read
+
+jobs:
+  Drainpipe-Deploy-Pantheon-Multidev:
+    if: contains(github.event.pull_request.labels.*.name, 'pantheon-multidev')
```
justafish, line 20:
```suggestion
    if: (contains(github.event.pull_request.labels.*.name, 'pantheon-multidev') && ${{ vars.PANTHEON_MULTIDEV_MANUAL }} == "true") || ${{ vars.PANTHEON_MULTIDEV_MANUAL }} != "true"
```

Could we do something like that instead of duplicating the files? (haven't tested it so not sure if it'll work!)

scaffold/github/workflows/PantheonReviewAppsManual.yml
```diff
@@ -0,0 +1,72 @@
+name: "Pantheon Review Apps"
+
+on:
+  pull_request:
+    types: [ labeled, unlabeled, opened, synchronize, reopened]
```
justafish, line 5:
```suggestion
    types: [ labeled, opened, synchronize, reopened]
```

justafish (changes requested):
Instead of duplicating these files could we

YesCT:
@mrdavidburns We don't want to use main on this branch. We want to use the release. Otherwise, we are running ahead of the release on DH.(And drainpipe won't match with drainpipe-dev, though I'm not clear on when that matters. @justafish or @davereid might be able to explain when it matters.)

mrdavidburns:
@YesCT Thanks for explaining why this pull request was not ready to be rebased with `main`. I went ahead and removed that commit from this branch. 

We will want to update this branch with the latest in `main` and confirm all checks passed before this is merged.

scaffold/github/workflows/PantheonReviewAppsManual.yml
YesCT: Moved to https://github.com/Lullabot/drainpipe/pull/670#discussion_r1734704482

scaffold/github/workflows/PantheonReviewAppsManual.yml
YesCT: committed in https://github.com/Lullabot/drainpipe/pull/670/commits/ab80299864a0713b181a1e19a3a0b9e7d35b018f

YesCT:
> I'm not quite following what went on there 😄 `drainpipe-dev` branches now get updated alongside branches opened in this repository, but there's no drainpipe-dev changes here
> 
> This branch needs updating with main and the changes requested addressed please 🙏

I moved the feedback to https://github.com/Lullabot/drainpipe/pull/670 

Please let's use that PR for development. (Sorry but I'm afraid to develop on this branch, while I'm using it on several customer repos.) I think we can avoid this in the future, by testing on a test site (lsm-examples), and not live customer sites. Part of my fear, is that when this branch were to be updated with main, our customer repos would still be on an old version of drainpipe-dev, and I'm not sure what troubles having those out of sync could cause. Part of my fear, is more simple, that we'd try pushing something here, that might temporarily interrupt our working build and the customer might be impacted.)

justafish:
@YesCT Whilst you could pin to a specific commit, I recommend you move your customer sites to using their own custom templates for this which aren't attached to Drainpipe, or your own fork - we can't really block active development on this branch and once it's merged it'll go away immediately

YesCT:
Yeah, next time I'll bring in unmerged code differently.

I made the other branch to continue development. https://github.com/Lullabot/drainpipe/pull/670

PR #552, opened 2024-04-29
Issue #369: Inject environment indicators into DDEV settings

### LINKED ISSUE #369: Generate Environment Indicator settings
Let's investigate automatically adding environment indicator settings, since we use it on all projects: https://architecture.lullabot.com/adr/20210609-environment-indicator/

geekygnr:
Fixes #369 

* Adds a way for environment indicator settings to be injected into ddev.
Not sure if this is the best approach but I followed how settings.php gets changed to include tugboat. I decided against generating a different settings.*.php file for ddev because I couldn't see a clean way to include it when needed and exclude it when its not.


justafish:
wfm! :+1: 

image

justafish (approved):

Issue #112, opened 2022-08-03
Separate out security checks from test:static


deviantintegral:
There can be good reasons to merge PRs that have open security issues:

1. There is a related upgrade that needs to be merged and tested first before upgrading the insecure package.
2. The site has some other issue going on concurrently ("the home page is down!").
3. If test:static is called from a pre-commit hook, it may be that another developer is handling the security update concurrently in a separate PR.

Let's remove test:security from test:static. On the one hand, I think it could be good to add test:security as a nightly CI job. However, with dependabot or renovate, that may not be useful.

justafish:
Whilst Renovate/Dependabot will pick up a security issue in the main branch, it doesn't tell you if the package you've just added to your PR contains a security issue (which has happened to me many times with node packages). I can understand it might be annoying for scenario #2 where time is of the essence, but isn't that more an issue of the required status checks you've enabled that's preventing it from being merged? Maybe a compromise here would be having a file of known security issues that can be skipped (as we've done in a previous project we both worked on :smile: )

deviantintegral:
I think a file of known issues to skip is good. It would be a small bump in development on open PRs ("why did my PR all of a sudden start failing"), but also one that's easy enough to fix.

hawkeyetwolf:
Do you have any particular format in mind (json?) or example for what the "file of known issues" would look like? Thanks! 🙏🏻

justafish:
Here's an example pulled from a project, but open to other suggestions!

```

# Format:
#
#  # Package name
#  # Title
#  # Reason ignored
#  Advisory URL

#####################
# Yarn Dependencies #
#####################

# url-regex
# Regex Denial of Service
# Only used for testing tools
https://www.npmjs.com/advisories/1550

#########################
# Composer Dependencies #
#########################
```

```
#!/bin/sh

yarn audit --level moderate
RESULT=$?

VULNERABLE=false
if [ "${RESULT}" != 0 ] && [ -f security-known-issues ]; then
  ADVISORIES=$(yarn audit --level moderate --json | jq -r 'select(.type=="auditAdvisory") | .data.advisory.url')
  for ADVISORY in ${ADVISORIES}; do
      grep -x -q ${ADVISORY} security-known-issues
      if [ $? ]; then
        echo "Known vulnerability: ${ADVISORY}"
      else
        echo "Unknown vulnerability: ${ADVISORY}"
        VULNERABLE=true
      fi
  done

[...]

PR #558, opened 2024-05-01
Separate test:security task out from test:static

davereid:
Fixes https://github.com/Lullabot/drainpipe/issues/112

How this affects consumers:

* Running `task test:static` will no longer run the `test:security` task by default.
* No scaffolded workflow changes

How this affects Drainpipe
* Adds a scheduled workflow for running the [_Test Static Tests_ workflow](https://github.com/Lullabot/drainpipe/actions/workflows/TestStatic.yml) day that will include the `task test:security` command as a best example.

### DIFF
--- a/.github/workflows/TestStatic.yml
+++ b/.github/workflows/TestStatic.yml
@@ -61,11 +61,11 @@ jobs:
         run: |
           sed -i '/^.*container_yamls.*$/i /** @phpstan-ignore-next-line */' web/sites/default/settings.php
 
-      - name: Run Security Tests
+      - name: Run security tests (only on scheduled runs)
         if: ${{ github.event_name == 'schedule' }}
         run: ddev task test:security
 
-      - name: Run Static Tests
+      - name: Run static tests
         run: ddev task test:static
 
       - name: Confirm custom directories are scanned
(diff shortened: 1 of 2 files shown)


.github/workflows/TestStatic.yml
```diff
@@ -5,6 +5,9 @@ on:
       - main
   pull_request:
     types: [opened, synchronize, reopened]
+  schedule:
+    - cron: '0 0 * * *'
```
deviantintegral, line 9: Is there an option for "random 24 hour period"? Or does github already load balance these things?

.github/workflows/TestStatic.yml
```diff
@@ -58,6 +61,10 @@ jobs:
         run: |
           sed -i '/^.*container_yamls.*$/i /** @phpstan-ignore-next-line */' web/sites/default/settings.php
 
+      - name: Run Security Tests
+        if: ${{ github.event_name == 'schedule' }}
+        run: ddev task test:security
```
deviantintegral, line 66: I wonder if we should update the name. Otherwise, reviewers will see "Run security tests" in regular PRs, and wonder why they're skipped?

.github/workflows/TestStatic.yml
davereid: I believe that GitHub already does their own load balancing on schedules, you'll often see them occur at off-minutes around the requested schedule.

justafish:
Thanks @davereid ! We discussed this today and decided to go with https://github.com/Lullabot/drainpipe/issues/112#issuecomment-2107854337

Give me a shout if you think that's not the direction we should go in and we can re-open this :+1:

Issue #220, opened 2023-06-29
npx @action-validator/cli is really slow compared to calling the binary directly


deviantintegral:
I noticed that the validator action is pretty slow. Here's examples from a repository with 34 files:

```console
$ time (find . -name '*yml' | xargs -n1 npx @action-validator/cli)

real	2m17.317s
user	1m11.874s
sys	0m13.308s
```

Calling the script directly (which is rust, so relatively fast!):

```console
$ time (find . -name '*yml' | xargs -n1 /mnt/ddev-global-cache/npm/_npx/7a1a696535fd56d2/node_modules/.bin/action-validator)

real	0m3.905s
user	0m6.540s
sys	0m1.423s
```

So, 2m14s of overhead just in spinning up nodejs and calling the script!

Since this is available as a single binary, I think we should download the binary directly, likely how we will do soon with task and local-security-checker.

deviantintegral:
Here's the Dockerfile lines I used to bring this into a project.

```
# Install action-validator for GitHub actions linting.
RUN /bin/bash -c '([[ $(arch) == "aarch64" ]] && curl -Lo /usr/local/bin/action-validator https://github.com/mpalmer/action-validator/releases/download/v0.5.2/action-validator_linux_arm64) || true'
RUN /bin/bash -c '([[ $(arch) == "x86_64" ]] && curl -Lo /usr/local/bin/action-validator https://github.com/mpalmer/action-validator/releases/download/v0.5.2/action-validator_linux_amd64) || true'
RUN chmod +x /usr/local/bin/action-validator
```

deviantintegral:
I've marked this as high priority since we're talking about a 35x slowdown for real-world projects.

PR #576, opened 2024-05-16
ISSUE-220: replace using npx to execute action-validator, use the binary instead

beto-aveiga:
### Description
- Issue: #220 
- Dowloads the binary of action-validator instead of using `npx`. Based on https://github.com/Lullabot/drainpipe/issues/220#issuecomment-1614915343
  - It took about 2 seconds to download the binary on my computer.
  - I tested that `action-validator` was executable with a new container, like: 
    - `docker run -it tugboatqa/php:8.2.15-apache-bullseye /bin/bash`


justafish (changes requested):
Thanks for the PR. This would download the binary each time the task is run, instead could you please change it to use the the same installation method as we use for task and the local security checker e.g.
https://github.com/Lullabot/drainpipe/blob/main/src/BinaryInstaller.php
https://github.com/Lullabot/drainpipe/blob/main/drainpipe-dev/src/DevBinaryInstallerPlugin.php

beto-aveiga:
@justafish I noticed that `action-validator` is only available for
- linux amd64
- linux arm64
- darwin amd64
- darwin arm64

Is that something we need to worry about? Or do something different in the binary installer? Or fallback to `npx` when the binary is not found?

beto-aveiga:
I just added a fallback in case the binary is not available.

deviantintegral:
> I just added a fallback in case the binary is not available.

That would be if you were running on Windows, directly on the host? Do we actually support doing that? @justafish thoughts?

If we keep it, I think we should document that it's specifically for Windows support as a comment.

tasks/github.yml
```diff
@@ -13,7 +13,13 @@ tasks:
...
           directories=".github/workflows .github/actions"
           if [ "{{ .directories }}" != "" ]; then directories="{{ .directories }}"; fi
-          find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \)  | xargs -n1 npx @action-validator/cli
+
+          if command -v action-validator &> /dev/null
+          then
+              find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \) | xargs -n1 action-validator
+          else
+              find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \) | xargs -n1 npx @action-validator/cli
+          fi
```
justafish, line 22:
```suggestion
          if [ -f ./vendor/bin/action-validator ]; then
              find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \) | xargs -n1 ./vendor/bin/action-validator
          else
              find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \) | xargs -n1 npx @action-validator/cli
          fi
```

tasks/github.yml
```diff
@@ -13,7 +13,13 @@ tasks:
       - |
           directories=".github/workflows .github/actions"
           if [ "{{ .directories }}" != "" ]; then directories="{{ .directories }}"; fi
-          find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \)  | xargs -n1 npx @action-validator/cli
+
+          if [ -f ./vendor/bin/action-validator ]; then
+              find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \) | xargs -n1 ./vendor/bin/action-validator
+          # Fallback for architectures that don't have a release e.g. Windows
+          else
```
justafish, line 20:
```suggestion
          # Fallback for architectures that don't have a release e.g. Windows
          else
```

tasks/github.yml
```diff
@@ -13,7 +13,14 @@ tasks:
       - |
           directories=".github/workflows .github/actions"
           if [ "{{ .directories }}" != "" ]; then directories="{{ .directories }}"; fi
-          find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \)  | xargs -n1 npx @action-validator/cli
+
+          if [ -f ./vendor/bin/action-validator ]; then
+              find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \) | xargs -n1 ./vendor/bin/action-validator
+          # Fallback for architectures that don't have a release e.g. Windows
+          # Fallback for architectures that don't have a release e.g. Windows
```
justafish, line 20:
```suggestion
```

justafish (approved):
I added a minor tweak to the explicit path of the binary, and added Andrew's suggestion for a comment. I've tested this locally with the drainpipe-test script and all looks to be working well :+1: 

Thanks all!

Issue #519, opened 2024-04-09
drupal:import-db and drupal:export-db don't compress data, though they use .sql.gz filenames


deviantintegral:
There's two related bugs that make imports and exports work, but in a confusing way.

First, the `export-db` command doesn't compress the database, but it saves it to `db.sql.gz`:

https://github.com/Lullabot/drainpipe/blob/47fb6f07df623cd0e29ee5cc1ccf45860f3c98d7/tasks/drupal.yml#L55

We need to either name the file to `.sql`, or do something like pass `--gzip`.

Likewise, the `import-db` command takes a `.sql.gz` file, but doesn't decompress it. So, if other code creates an expected compressed file, it will fail.

https://github.com/Lullabot/drainpipe/blob/47fb6f07df623cd0e29ee5cc1ccf45860f3c98d7/tasks/drupal.yml#L47

Normally, I'd suggest that we should use a more modern compression program, but both Acquia and Pantheon download in gzip format, and having mixed compression formats would be tricky for devs to manage.

mrdavidburns:
@alexanderpatriciop See if `pigz` will work w/ Drainpipe for compressed databases. If so update the pull request. If not, leave a comment on the pull request and then pass it over to me for review.

PR #581, opened 2024-05-27
Issue #519: Fix drupal:import-db and drupal:export-db don't compress data.

alexanderpatriciop:
### Relates to:
- #519 

### Description:
- Fix drupal:import-db and drupal:export-db don't compress data.

### DIFF
--- a/tasks/drupal.yml
+++ b/tasks/drupal.yml
@@ -44,7 +44,8 @@ tasks:
       - echo "🚮 Dropping existing database"
       - ./vendor/bin/drush {{ .site }} sql:drop --yes
       - echo "📰 Importing database"
-      - gunzip < $DB_DIR/db.sql.gz | ./vendor/bin/drush {{ .site }} sql:cli
+      - gunzip -k $DB_DIR/db.sql.gz
+      - ./vendor/bin/drush {{ .site }} sql:query --file=$DB_DIR/db.sql
       - defer: rm -f $DB_DIR/db.sql
   export-db:
     desc: "Exports a database fetched with a *:fetch-db command"
--- a/tasks/drupal.yml
+++ b/tasks/drupal.yml
@@ -41,12 +41,10 @@ tasks:
     env:
       DB_DIR: '{{default "/var/www/html/files/db" .DB_DIR}}'
     cmds:
-      - cp $DB_DIR/db.sql.gz $DB_DIR/db-copy.sql.gz
       - echo "🚮 Dropping existing database"
       - ./vendor/bin/drush {{ .site }} sql:drop --yes
       - echo "📰 Importing database"
-      - ./vendor/bin/drush {{ .site }} sql:query --file-delete --file=$DB_DIR/db.sql.gz
-      - mv $DB_DIR/db-copy.sql.gz $DB_DIR/db.sql.gz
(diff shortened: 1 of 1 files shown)


alexanderpatriciop:
On the env that I tested the task drupal:import-db after importing deletes the .gz file if it is compressed, so I made a copy to preserve the file. (also I added the --file-delete parameter because in other environments it may not clear itself)

alexanderpatriciop:
@justafish Thank you for the feedback, you are right, it's the same bug, I added the workaround suggested on the last commit, however, I realized it has an impact on the performance (at least on my local environment) because with the first approach (copy the .gz file), for importing the database it takes around 40 seconds, but with the latest approach it takes around 4 minutes.
We may need to revert the last commit. Please, let me know your thoughts.

deviantintegral:
That's very odd to see such a difference using a pipe like that. Do you have testing or reproduction steps others can use to check?

alexanderpatriciop:
- Drupal => 10.2.6
- lullabot/drainpipe => ^3.8.0
- drush/drush => ^12.5.2

Apply the patch https://github.com/Lullabot/drainpipe/commit/3b3bb598b96bdf73deb00d46c0900903c4337926.diff

* Run the commands
  * `ddev task drupal:export-db`
  * `ddev task drupal:import-db`

Remove the before patch and apply https://patch-diff.githubusercontent.com/raw/Lullabot/drainpipe/pull/581.diff

* Run the commands
  * `ddev task drupal:export-db`
  * `ddev task drupal:import-db`

And compare the execution time.

tasks/drupal.yml
```diff
@@ -44,15 +44,15 @@ tasks:
       - echo "🚮 Dropping existing database"
       - ./vendor/bin/drush {{ .site }} sql:drop --yes
       - echo "📰 Importing database"
-      - ./vendor/bin/drush {{ .site }} sql:query --file=$DB_DIR/db.sql.gz
+      - gunzip < $DB_DIR/db.sql.gz | ./vendor/bin/drush {{ .site }} sql:cli
```
mrdavidburns, line 47: A few years ago @m4olivei suggested using pigz instead of gunzip because it had significant performance benefits. 

https://www.clouvider.com/knowledge_base/compression-options-in-linux-gzip-vs-pigz/

m4olivei: Ohh, thanks for linking that article. I never did learn the reason _why_ `pigz` was faster, just that it was 💡

deviantintegral: I don't think pigz is in the default ddev image, so we'd need to add a Dockerfile to install it. Which IMO is fine, we do it on most of our projects anyways.

I do wonder if it's worth asking about adding it to the default web images in ddev upstream.

alexanderpatriciop: @mrdavidburns 
I tried `pigz` with the following command:
* `pigz -dc file.gz`

But the time execution was more than 2 minutes, so I unzipped the file before importing with `sql:query` and it worked fine, it takes the same time execution with `pigz` and `gunzip` so I left `gunzip` because the other needed to add a Dockerfile to install it as @deviantintegral mentioned

mrdavidburns: @alexanderpatriciop Were you trying pigz with an existing project or just a fresh site install?

alexanderpatriciop: Yes, I tried it with an existing project

mrdavidburns (approved):

PR #582, opened 2024-05-27
ISSUE-318: composer-lock-diff descriptions with \r\n

beto-aveiga:
### Description
- Uses `jq` to format `composer-lock-diff` results.
- Updates the expected JSON files for the `composer-lock-dff` tests.
- Replaces `\n` with `\r\n` for new lines.
- Uses `\r\n` instead of `\n` wherever possible.

From issue #318


alexanderpatriciop (approved):
LGTM!

justafish:
Fixed in #636

PR #595, opened 2024-06-19
WIP Renovate/php 8.x  luisaveiga

beto-aveiga:
## Description
This PR is a modification of #588 

## Background
The tugboat is not working. If I recall correctly, I saw a quota issue explaining why the tests are failing.

Do you know if this is what is needed? I added restrictions to run tests with PHP 8.3 only.
I also set the default DDEV version to 1.23.2 because I couldn't replicate the issues on my machine when I started. DDEV v1.23.2 already has PHP 8.3.

There are PHP version strings on the Drainpipe Tugboat still, but I prefer to first have feedback if this is the correct approach. If so, I can update the missing ones.

### DIFF
--- a/drainpipe-dev/composer.json
+++ b/drainpipe-dev/composer.json
@@ -10,7 +10,7 @@
         "psr-4": {"Lullabot\\DrainpipeDev\\Tests\\Functional\\": "tests/src/functional/"}
     },
     "require": {
-        "php": "^8.1",
+        "php": "^8.3.8",
         "ext-json": "*",
         "composer-plugin-api": "^2.0",
         "phpunit/phpunit": "^9.6.19",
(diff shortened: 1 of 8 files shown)


drainpipe-dev/composer.json
```diff
@@ -10,7 +10,7 @@
         "psr-4": {"Lullabot\\DrainpipeDev\\Tests\\Functional\\": "tests/src/functional/"}
     },
     "require": {
-        "php": "^8.1",
+        "php": "^8.3.8",
```
mrdavidburns, line 13: We're going to want to support PHP 8.1, 8.2, and 8.3.

beto-aveiga: I'm unclear about what to do next.
If I don't change the PHP version, it won't work.
Do you have a suggestion on how to do it? My approach may be wrong.

mrdavidburns:
@deviantintegral @justafish Is Drainpipe able to support multiple point releases of PHP? I thought it supported 8.1 and 8.2 right now, but I could be wrong. 

Suggestions on how we can support 8.1, 8.2, and 8.3?

mrdavidburns:
Fixed in main branch.

Issue #605, opened 2024-07-10
Fix Tugboat  `sessions table doesn't exist`


mrdavidburns:
All Tugboat builds on recent pull requests are failing with this error.

```
[sync] 🧹 Sanitising database
[sync] The following operations will be performed:
[sync]  * Remove comment display names and emails.
[sync]  * Truncate sessions table.
[sync]  * Sanitize text fields associated with users.
[sync]  * Sanitize user passwords.
[sync]  * Sanitize user emails.
[sync]  * Preserve user emails and passwords for the specified roles.
[sync] 
[sync]  // Do you want to sanitize the current database?: yes.                         
[sync] 
[sync] 
[sync] In ExceptionHandler.php line 56:
[sync]                                                                                
[sync]   SQLSTATE[42S02]: Base table or view not found: 1146 Table 'tugboat.sessions  
[sync]   ' doesn't exist: TRUNCATE "sessions"; Array                                  
[sync]   (                                                                            
[sync]   )                                                                            
[sync]                                                                                
[sync] 
[sync] In StatementWrapperIterator.php line 113:
[sync]                                                                                
[sync]   SQLSTATE[42S02]: Base table or view not found: 1146 Table 'tugboat.sessions  
[sync]   ' doesn't exist                                                              
[sync]                                                                                
[sync] 
task: Failed to run task "sync": exit status 1
```

This was not happening with any of the recent pull requests that have been merged into the `main` branch.
* https://github.com/Lullabot/drainpipe/pulls?q=is%3Apr+is%3Aclosed+is%3Amerged

mrdavidburns:
I pushed an empty branch and created a pull request. It failed then I addressed some messages I saw when running `composer install` and pushed it to the same branch and it still fails with same error.

* https://github.com/Lullabot/drainpipe/pull/603

beto-aveiga:
Here is a PR where the checks are passing.
https://github.com/Lullabot/drainpipe/pull/607

I updated it based on the conversation here: https://github.com/Lullabot/drainpipe/pull/603

mrdavidburns:
@beto-aveiga What if we temporarily pin a specific version of Drupal 10.2.x in the Tugboat build?

https://github.com/Lullabot/drainpipe/blob/main/.tugboat/steps/1-init.sh#L77

Then we can unpin the version once Drush creates a stable release with the necessary fix?

PR #607, opened 2024-07-11
PHP ^8.2 and Drush 13 dev to allow Tugboat builds

beto-aveiga:
- Updates to PHP 8.2 (at least)
- Updates to drush/drush 13 (dev)

Related to: https://github.com/Lullabot/drainpipe/issues/605

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,14 +28,15 @@
             "\\Lullabot\\Drainpipe\\BinaryInstallerPlugin"
         ],
         "drupal-scaffold": {
-            "gitignore": true,
+            "gitignore": "true",
             "file-mapping": {
                 "[project-root]/.env": {
                     "mode": "replace",
                     "path": "scaffold/env/env",
                     "overwrite": false
                 }
-            }
+            },
+            "allowed-packages": ["lullabot/drainpipe", "lullabot/drainpipe-dev"]
         }
     },
     "config": {
--- a/composer.json
+++ b/composer.json
@@ -10,10 +10,10 @@
         "psr-4": {"Lullabot\\Drainpipe\\Tests\\Functional\\": "tests/src/functional/"}
     },
     "require": {
-        "php": "^8.1",
+        "php": "^8.2",
         "ext-json": "*",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -10,10 +10,10 @@
         "psr-4": {"Lullabot\\Drainpipe\\Tests\\Functional\\": "tests/src/functional/"}
     },
     "require": {
-        "php": "^8.1",
+        "php": "^8.2",
```
mrdavidburns, line 13: Looks like PHP 8.1 is still supported until Jan. 1 2026.
* https://www.php.net/supported-versions.php

Wouldn't `^8.1`  still allow PHP 8.2 to be installed?
Why do we need to force a minimum of PHP 8.2?

composer.json
```diff
@@ -10,10 +10,10 @@
...
         "ext-json": "*",
         "composer-plugin-api": "^2.0",
-        "drush/drush": "^11|^12|^13",
+        "drush/drush": "13.x-dev",
```
mrdavidburns, line 16: We definitely need Drainpipe to support 11, 12, and 13. 
We can't force all projects to dev version of 13. But I understand we needed to do that to get Tugboat's `drush sql-sanitize` to work because of https://www.drupal.org/node/3431286

mrdavidburns: @deviantintegral @justafish knowing that we can't (nor do we want to) force everyone to PHP 8.2 and a dev release of Drush 13, what are our options here to unblock pull requests that are only failing on Tugboat builds with this sessions table error?

deviantintegral: I think we should reach out to Moshe and ask about rolling a Drush 12 point release. Since this breaks for drupal 10.3 (and not 11) I think getting that out is the best thing?

mrdavidburns: @deviantintegral to unblock our recent PRs would it be okay to pin the Drupal version while we wait for Drush to create new releases? https://github.com/Lullabot/drainpipe/pull/608

mrdavidburns:
Closing in favor of #608

Issue #528, opened 2024-04-15
Allow DDEV binary to be overridden


justafish:
See https://github.com/Lullabot/drainpipe/pull/453/files#r1501209503

Issue #175, opened 2023-02-15
task not being installed, dotenv.php missing from autoload-dev


mrdavidburns:
Seeing this error appear when running `ddev composer install --no-cache` after updating to 1.3.1

```
[Drainpipe] vendor/lullabot/drainpipe/scaffold/env/dotenv.php' missing from autoload-dev files

In Filesystem.php line 254:

  /dev/null/files/lullabot/drainpipe/bin/task/3.18.0 does not exist and could not be created.
```

Note: this is happening on many, but not all, pull requests and builds on Tugboat.
I have been able to replicate locally by running:
`ddev composer install --no-cache`

possibly related: #172 and #165

quicksketch:
I'm seeing this same problem on Evergreen.

mrdavidburns:
If it's still there try removing `vlucas/phpdotenv` and update `composer.json` with the correct autoload from drainpipe README.md

quicksketch:
A little more information: If you run `ddev composer install --no-cache -vvv` you can get a more detailed explanation of the failure:

```

Exception trace:
  at phar:///usr/local/bin/composer/src/Composer/Util/Filesystem.php:254
 Composer\Util\Filesystem->ensureDirectoryExists() at /var/www/html/vendor/lullabot/drainpipe/src/BinaryInstaller.php:200
 Lullabot\Drainpipe\BinaryInstaller->installBinary() at /var/www/html/vendor/lullabot/drainpipe/src/BinaryInstaller.php:168
 Lullabot\Drainpipe\BinaryInstaller->installBinaries() at /var/www/html/vendor/lullabot/drainpipe/src/BinaryInstaller.php:122
 Lullabot\Drainpipe\BinaryInstaller->onPostInstallCmd() at phar:///usr/local/bin/composer/src/Composer/EventDispatcher/EventDispatcher.php:200
 Composer\EventDispatcher\EventDispatcher->doDispatch() at phar:///usr/local/bin/composer/src/Composer/EventDispatcher/EventDispatcher.php:123
 Composer\EventDispatcher\EventDispatcher->dispatchScript() at phar:///usr/local/bin/composer/src/Composer/Installer.php:374
 Composer\Installer->run() at phar:///usr/local/bin/composer/src/Composer/Command/InstallCommand.php:144
 Composer\Command\InstallCommand->execute() at phar:///usr/local/bin/composer/vendor/symfony/console/Command/Command.php:298
[...]

PR #619, opened 2024-07-22
Try to avoid 'Unable to add newly converted phar... to the list of ph…

beto-aveiga:
…ars, a phar with that name already exists' due to https://bugs.php.net/bug.php\?id\=75101 [skip ci]


mrdavidburns:
@beto-aveiga would this be related to any of these issues, just so we associate this with an issue that we're tracking.

* #386
* #135
* #175 
* #528
* #208

src/BinaryInstaller.php
```diff
@@ -206,7 +206,7 @@ protected function installBinary($binary, $version, $url, $sha, $hashalgo = 'sha
         else {
             $cacheFolder = sys_get_temp_dir() . \DIRECTORY_SEPARATOR . $binary . \DIRECTORY_SEPARATOR . $version;
         }
-        $cacheDestination = $cacheFolder . \DIRECTORY_SEPARATOR . $fileName;
+        $cacheDestination = $cacheFolder. \DIRECTORY_SEPARATOR . bin2hex(random_bytes(5)) . $fileName;
```
deviantintegral, line 209: Won't this mean we effectively invalidate the cache every time, and never use it?

beto-aveiga: Yes. I had to do this because there was an error building Tugboat.
This error happened while running the composer install.
I couldn't find any other way.

mrdavidburns: Would it make sense to only do the random string cache file after we recognize the error that is causing us to do this. This way we still get the benefits of cache for instances where everything is working fine?

deviantintegral:
I think this is fixed now, and we know that this PR doesn't work as is because it just breaks the cache on every execution. Closing!

Issue #616, opened 2024-07-19
Fix Pantheon deploy error


mrdavidburns:
Relates to: #601 

Pantheon DEV deploy still failing on `update` step with v3.8.1
Looks like the update command runs but it reports back as failed.

```
Failed to execute command ./vendor/bin/task -l | grep '* update: ': exit status 1
[drupal:update]  [success] No pending updates.
[drupal:update]  [success] Cache rebuild complete.
Notice: [drupal:update]  [notice] There are no changes to import.
Notice: [drupal:update]  [notice] There are no changes to import.
[drupal:update]  [success] Cache rebuild complete.
[drupal:update]  [success] No pending deploy hooks.
[drupal:update] Config export does not match database.
Notice: [drupal:update]  [notice] No differences between DB and sync directory.
task: Failed to run task "drupal:update": exit status 1
Failed to execute command ./vendor/bin/task drupal:update site=@***.pr-3836: exit status 201
Error: Process completed with exit code 1.
```

mrdavidburns:
Making note that this may be caused by a change from Pantheon. The DEV deploy actions worked fine and did not change for a few projects then suddenly started having these types of errors.

vordude:
Hi Matt,

Arvin here, with the Customer Success team at Pantheon.

We've received your issue and the team was able to look into it. With the logs that you provided, we may need further clarification to help us understand what's happening in these GitHub actions from the logs and what the problem is with the site.

Mainly on this log here:

[...]

PR #622, opened 2024-07-23
Issue #616 - Fix Pantheon deploy error

mrdavidburns:
fixes: #616 

This patch created by @alexanderpatriciop has been tested on multiple projects and does complete the deployment without failure

### DIFF
--- a/tasks/drupal.yml
+++ b/tasks/drupal.yml
@@ -68,7 +68,8 @@ tasks:
       - ./vendor/bin/drush {{.site}} --yes cache:rebuild
       - ./vendor/bin/drush {{.site}} --yes deploy:hook
       - |
-        if [[ $(./vendor/bin/drush {{.site}} config:status --format=json --state=Different) != '[]' ]]; then
+        config_status_output=$(./vendor/bin/drush {{.site}} config:status --format=json --state=Different | tail -n1)
+        if [[ $config_status_output != '[]' ]]; then
           echo "Config export does not match database."
           ./vendor/bin/drush {{.site}} config:status
           exit 1;
--- a/tasks/drupal.yml
+++ b/tasks/drupal.yml
@@ -68,6 +68,8 @@ tasks:
       - ./vendor/bin/drush {{.site}} --yes cache:rebuild
       - ./vendor/bin/drush {{.site}} --yes deploy:hook
       - |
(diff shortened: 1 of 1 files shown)


tasks/drupal.yml
```diff
@@ -68,7 +68,8 @@ tasks:
       - ./vendor/bin/drush {{.site}} --yes cache:rebuild
       - ./vendor/bin/drush {{.site}} --yes deploy:hook
       - |
-        if [[ $(./vendor/bin/drush {{.site}} config:status --format=json --state=Different) != '[]' ]]; then
+        config_status_output=$(./vendor/bin/drush {{.site}} config:status --format=json --state=Different | tail -n1)
```
deviantintegral, line 71: So the only real difference here is `tail -n1`? Is drush outputting some text beyond the json output? I would have expected any non-json text to be in stderr, not stdout.

mrdavidburns: From what I've seen by running the commands myself the last characters are `[]`. However, we have merged this patch on a project and it completes without error.

See my comment here about the output: https://github.com/Lullabot/drainpipe/issues/616#issuecomment-2243987592

penyaskito:
```suggestion
       # drush config:status --format=json is outputting notices in Pantheon even with the json format, 
       # so we need to tail the last line.
        config_status_output=$(./vendor/bin/drush {{.site}} config:status --format=json --state=Different | tail -n1)
```
Let's at least add a brief comment for this.
This has worked for our project dev deployments, which were failing before.

penyaskito: @deviantintegral This is executed with command substitution as ' $(./vendor/bin/drush {{.site}} config:status --format=json --state=Different)`
As far as I know that doesn't include stderr, but I'm not 100% sure.

deviantintegral: It doesn't!

```
$ ls $(echo composer.lock 1>&2)
composer.lock
README.md     composer.json  config             node_modules  phpcs.xml.dist  postcss.config.js  scripts             sites.json  vendor  yarn.lock
Taskfile.yml  composer.lock  eslint.config.mjs  package.json  phpstan.neon    renovate.json      sites-domains.json  test        web
```

If it did, the above would have only shown composer.lock. Makes sense, thanks for the TIL'ed!

deviantintegral (approved):

PR #624, opened 2024-07-30
Update documentation for Tugboat [skip ci]

beto-aveiga:
Updates documentation for Tugboat
The filename to override values in the Tugboat config file is wrong.
Updated from config.drainpipe.yml to config.drainpipe-override.yml

See https://github.com/Lullabot/drainpipe/blob/db84b66a2a5ce7aa0cec5b30f4f0e834d6bc9cc6/src/ScaffoldInstallerPlugin.php#L353


mrdavidburns (approved):
Thank you!

Issue #597, opened 2024-06-26
Add GitHub Action to run `composer outdated` and `composer audit`


mrdavidburns:
We have noticed on some projects that Renovate is not picking up major version upgrades or we have pinned versions that are not getting new releases. We would like to have a GitHub action that we can run on a weekly or monthly basis so that anyone with access to the logs of that job can identify these packages then we can create tickets to get those packages updated.

mrdavidburns:
@justafish Yes, and we do have that configured for things like Drupal core major and point releases. 
I created this when reviewing all our LSM projects and was running those 2 commands to determine if anything was being missed. It doesn't take long to checkout the latest code and run these commands manually. I was thinking this could be helpful to avoid having to do that step, but agree it probably isn't needed for all projects using Drainpipe.

Closing.

YesCT:
There have been cases where renovate did not alert us. I'll share a slack link in our internal devops channel.

Issue #283, opened 2023-10-06
Faster composer_lock_diff without DDEV


beto-aveiga:
Not using DDEV to generate composer-lock-diff information was much faster in our tests.

### Time comparison (with and without DDEV)

The difference was from 4 minutes to 30 seconds.

![image](https://github.com/Lullabot/drainpipe/assets/3916979/012eeecf-e425-4ac4-89c5-2111d0c790b4)

### The code

This is the code we are using for that workflow:

```
name: "Composer Lock Diff"

on:
  pull_request:
    branches:
      # List target branches, not source branches.
      - '*'

concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
  cancel-in-progress: true

jobs:
  Composer-Lock-Diff:
    # Lock Ubuntu to always get PHP 8.1
    # See: https://github.com/shivammathur/setup-php#github-hosted-runners
    runs-on: ubuntu-22.04
    permissions:
      contents: read
      pull-requests: write

    steps:
      - uses: actions/checkout@v3
        with:
          ref: ${{ github.event.pull_request.base.ref }}

      - uses: actions/checkout@v3
        with:
          fetch-depth: 2

      - name: Set environment variables
        run: |
          echo "PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')" >> $GITHUB_ENV

      - name: Install Composer
        run: |
          php --run "copy('https://getcomposer.org/installer', 'composer-setup.php');"
          php composer-setup.php --filename=composer --version=2.5.8
          php --run "unlink('composer-setup.php');"
          chmod +x composer
          mv composer /usr/local/bin/composer

      - name: Install dependencies
        run: composer install

      - name: Install composer lock diff
        run: composer global require davidrjonas/composer-lock-diff:^1.0

      - name: Get the ID of the last review (if any)
        run: |

          response=$(curl --location \
            --header "Accept: application/vnd.github+json" \
            --header "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}"\
            --header "X-GitHub-Api-Version: 2022-11-28" \
            https://api.github.com/repos/${{ github.repository }}/pulls/$PR_NUMBER/reviews)

          # LAST_REVIEW_ID won't be empty when we already have a review.
          # If so, we will update the review later, instead of creating one.
          github_action_reviews=$(echo "$response" | jq '.[] | select(.user.login == "github-actions[bot]")')
[...]

PR #636, opened 2024-08-07
Issue #112 #332: Seperate Security Checks and Replace Composer Lock Diff

### LINKED ISSUE #112: Separate out security checks from test:static

justafish:
Example: https://github.com/justafish/test-drainpipe/pull/1#discussion_r1706885265

Fixes #112, some of #597, #332, #318, and #283


composer.json
```diff
@@ -19,8 +19,8 @@
         "vlucas/phpdotenv": "^4|^5"
     },
     "require-dev": {
-        "composer/composer": "^2.7.7",
```
deviantintegral, line 22: Do we need to disable bumping in renovate?

scaffold/github/workflows/Security.yml
```diff
@@ -0,0 +1,86 @@
...
+
+      - name: Build Project
+        run: ddev composer install
+
+      - name: Run Security Checks
+        run: ddev task security composer_lock_diff_from=${{ github.base_ref }} composer_lock_diff=composer-lock-diff.md
+
+      - name: Publish Results
+        run: |
+          COMPOSER_LOCK_DIFF="<!-- composer-lock-diff -->
```
deviantintegral, line 51: Can we make this at least a shell command or a task, so we can run and test this outside of GitHub actions?

scaffold/github/workflows/Security.yml
```diff
@@ -0,0 +1,86 @@
+name: "Security"
```
deviantintegral, line 1: Forgive me if I’ve just missed the conversation, but last we discussed we thought we’d include this in the static check job to avoid the additional CI spin up costs of a separate ddev runner.

deviantintegral: Oh I misread this, we are at least combining two jobs together. 👍

scaffold/github/workflows/Security.yml
justafish: This part is all very GitHub actions specific and relies on variables set in the environment, so doesn't really make sense to be in a task

composer.json
justafish: maybe, I need to look into this some more - I wasn't able to install drainpipe-dev following our installation instructions with a fresh copy of Drupal. Just wanted to get the PR up in the meantime for some feedback.

justafish: this should be resolved now

README.md
```diff
@@ -391,7 +391,22 @@ They are composite actions which can be used in any of your workflows e.g.
 Tests can be run locally with [act](https://github.com/nektos/act):
 `act -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:runner-latest -j Static-Tests`
 
-### Composer Lock Diff
+### Security
+```json
+"extra": {
+  "drainpipe": {
+    "github": ["Security"]
```
mrdavidburns, line 398: Nice!

mrdavidburns (approved):

Issue #638, opened 2024-08-08
Installation is broken


justafish:
Following our installation instructions or using the test script:

```
+ ddev composer require 'lullabot/drainpipe-dev @dev' --dev --with-all-dependencies
./composer.json has been updated
Running composer update lullabot/drainpipe-dev --with-all-dependencies
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - phpunit/phpunit[9.6.20, ..., 9.6.x-dev] require sebastian/diff ^4.0.6 -> found sebastian/diff[4.0.6, 4.0.x-dev] but these were not loaded, likely because it conflicts with another require.
    - lullabot/drainpipe-dev dev-main requires phpunit/phpunit ^9.6.20 -> satisfiable by phpunit/phpunit[9.6.20, 9.6.x-dev].
    - Root composer.json requires lullabot/drainpipe-dev @dev -> satisfiable by lullabot/drainpipe-dev[dev-main].


Installation failed, reverting ./composer.json and ./composer.lock to their original content.
Composer [require lullabot/drainpipe-dev @dev --dev --with-all-dependencies] failed, composer command failed: exit status 2. stderr=
```

Issue #615, opened 2024-07-18
Fix linting for Storybook


mrdavidburns:

elvism-lullabot:
@mrdavidburns @deviantintegral 
not sure if this branch/commit could help here https://github.com/Lullabot/drainpipe/commit/0b8aa9b119e16cbe818e425605db19d079e520ba

deviantintegral:
@elvism-lullabot we have a PR open at https://github.com/Lullabot/drainpipe/pull/613 for this.

I think both are substantially similar, though the PR handles any docroot (and not just web) and is conditional within the same command.

PR #639, opened 2024-08-08
Issue #638: Fix all our tests

justafish:
- Replace sserbin/twig-linter with vincentlanglet/twig-cs-fixer/ - sserbin/twig-linter depends on an old version of symfony/console and core is adopting this new package at some point https://www.drupal.org/project/drupal/issues/3284817 - it also has a much better extension mechanism which will help if something like #615 comes up again
- Update Pantheon settings for Drupal 11
- Update PHPUnit configurations and tests for Drupal 11
- Fix mysterious yarn issues with DDEV

### DIFF
--- a/.github/workflows/TestMetapackages.yml
+++ b/.github/workflows/TestMetapackages.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Create a Drupal project
-        run: composer create-project drupal/recommended-project . --ignore-platform-req=ext-gd
+        run: composer create-project drupal/recommended-project . --ignore-platform-reqs
 
       - uses: actions/checkout@v4
         with:
@@ -35,7 +35,7 @@ jobs:
       - name: Setup Project
         run: |
           ddev config --auto
-          ddev config --nodejs-version "18"
+          ddev config --nodejs-version "22"
           ddev start
           ddev composer config extra.drupal-scaffold.gitignore true
           ddev composer config --json extra.drupal-scaffold.allowed-packages '["lullabot/drainpipe-dev", "lullabot/drainpipe"]'
@@ -197,7 +197,6 @@ jobs:
 
       - name: Setup Sass
         run: |
-          ddev config --nodejs-version=21
           ddev restart
           ddev yarn set version berry
           ddev yarn set version 4
@@ -344,7 +343,6 @@ jobs:
 
       - name: Setup Sass
         run: |
-          ddev config --nodejs-version=21
           ddev restart
           ddev yarn set version berry
           ddev yarn set version 4
--- a/.github/workflows/TestMetapackages.yml
+++ b/.github/workflows/TestMetapackages.yml
@@ -129,6 +129,7 @@ jobs:
         run: |
           ddev exec corepack disable
           ddev exec npm i -g yarn@1.22.1
+          ddev exec 'ln -s $(npm root -g)/yarn/bin/yarn /usr/local/bin/yarn'
           ddev exec yarn init -y
           ddev exec yarn add file:./drainpipe/metapackages/sass --dev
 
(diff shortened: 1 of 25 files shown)


.github/workflows/TestPHPUnit.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         uses: GuillaumeFalourd/assert-command-line-output@v2.3
         with:
           command_line: xmllint --xpath 'string(//testsuites/testsuite/@tests)' test_result/phpunit-functional.xml
-          contains: 18
+          contains: 6
```
justafish, line 81: These were previously also running the unit tests apparently

.github/workflows/TestPHPUnit.yml
```diff
@@ -147,7 +147,7 @@ jobs:
         uses: GuillaumeFalourd/assert-command-line-output@v2
         with:
           command_line: xmllint --xpath 'string(//testsuites/testsuite/@tests)' test_result/phpunit-functional.xml
-          contains: 21
+          contains: 9
```
justafish, line 150: These were previously also running the unit tests apparently

deviantintegral:
You've buried the lede here with "Drupal 11 compatibility"!

.github/workflows/TestMetapackages.yml
```diff
@@ -271,10 +273,12 @@ jobs:
...
         run: |
-          ddev yarn set version classic
-          ddev yarn init -y
-          ddev yarn add focus-trap@^6.7.3
-          ddev yarn add file:./drainpipe/metapackages/javascript --dev
+          ddev exec corepack disable
+          ddev exec npm i -g yarn@1.22.1
+          ddev exec 'ln -s $(npm root -g)/yarn/bin/yarn /usr/local/bin/yarn'
+          ddev exec yarn init -y
+          ddev exec yarn add focus-trap@^6.7.3
```
deviantintegral, line 280: what's this package for? Is it just an easy way to do a test?

.github/workflows/test-production-build.yml
```diff
@@ -12,7 +12,11 @@ jobs:
     strategy:
       matrix:
         php-version: [8.1, 8.2, 8.3]
-
+        include:
+          - php-version: 8.1
+            drupal-version: ":^10"
```
deviantintegral, line 17: nice.

tests/fixtures/phpunit/testmodule/modules/testsubmodule/tests/src/Functional/UserCreateTest.php
```diff
@@ -37,9 +37,10 @@ class UserCreateTest extends BrowserTestBase {
     public function testUserAdd() {
         $user = $this->drupalCreateUser(['administer users']);
         $this->drupalLogin($user);
+        $request_time = \Drupal::time()->getRequestTime();
 
-        $this->assertEquals(REQUEST_TIME, $user->getCreatedTime(), 'Creating a user sets default "created" timestamp.');
-        $this->assertEquals(REQUEST_TIME, $user->getChangedTime(), 'Creating a user sets default "changed" timestamp.');
+        $this->assertEquals($request_time, $user->getCreatedTime(), 'Creating a user sets default "created" timestamp.');
+        $this->assertEquals($request_time, $user->getChangedTime(), 'Creating a user sets default "changed" timestamp.');
```
deviantintegral, line 43: 👏

deviantintegral (approved):
I wonder if we should consider this a BC break, and bump Drainpipe to v4? In particular, if the twig linter has any sort of different rules, or if teams were expecting the twig linter in other custom tasks.

.github/workflows/TestMetapackages.yml
justafish: It's testing whether the package import works correctly when bundling

justafish:
@deviantintegral I've made some tweaks to make it compatible with both D10 and D11. I don't think we need to bump to v4 as the twig linter wasn't configurable previously.

Issue #508, opened 2022-07-05
Have the scaffold create example GitHub actions for Static and Functional tests


justafish:

mrdavidburns:
Drainpipe may already be doing this. Please confirm and comment here.

PR #652, opened 2024-08-15
Issue #508: Add templates for static and functional tests

justafish:


justafish:
This will need a bit more work to support getting secrets into environment variables, but will do as a first pass

PR #654, opened 2024-08-15
Fix Composer Lock Diff workflow for PRs that don't have composer.lock…

justafish:
… changes


mrdavidburns (approved):
Yes please and thank you!

Issue #655, opened 2024-08-15
Don't run static checks on contrib


YesCT:
Don't run static checks on contrib.

(Also should not be running on vendor and core, but that might already be as desired.)

YesCT:
Related: https://github.com/Lullabot/drainpipe/issues/508

bronzehedwick:
This issue is preventing a project upgrade.

mrdavidburns:
Here's a project that had 4 contrib modules failing static checks.

```
[test:lint]  [OK] All 1 YAML files contain valid syntax.                                    
[test:lint] 
[test:lint]  KO web/modules/contrib/field_group/templates/field-group-html-element.html.twig
[test:lint]  ------- ----------------------------------------------------- 
[test:lint]   ERROR   25   | {% endif %}                                   
[test:lint]           26   | {{children}}                                  
[test:lint]           >>   | Expecting 1 whitespace after "{{"; found 0.   
[test:lint]           27   | {% if collapsible %}                          
[test:lint]  ------- ----------------------------------------------------- 
[test:lint]   ERROR   25   | {% endif %}                                   
[test:lint]           26   | {{children}}                                  
[test:lint]           >>   | Expecting 1 whitespace before "}}"; found 0.  
[test:lint]           27   | {% if collapsible %}                          
[test:lint]  ------- ----------------------------------------------------- 
[test:lint]   ERROR   29   | {% endif %}                                   
[test:lint]           30   | </{{ wrapper_element }}>                      
[test:lint]           >>   | A file must end with 1 blank line; found 0    
[test:lint]  ------- ----------------------------------------------------- 
[test:lint] 
[test:lint]  KO web/modules/contrib/field_group/modules/field_group_accordion/templates/field-group-accordion-item.html.twig
[test:lint]  ------- ------------------------------------------------------------------------- 
[test:lint]   ERROR   37   |   {% if description %}<div class="description"></div>{% endif %}  
[test:lint]           38   |   {{children}}                                                    
[test:lint]           >>   | Expecting 1 whitespace after "{{"; found 0.                       
[test:lint]           39   | </div>                                                            
[test:lint]  ------- ------------------------------------------------------------------------- 
[test:lint]   ERROR   37   |   {% if description %}<div class="description"></div>{% endif %}  
[test:lint]           38   |   {{children}}                                                    
[test:lint]           >>   | Expecting 1 whitespace before "}}"; found 0.                      
[test:lint]           39   | </div>                                                            
[...]

PR #656, opened 2024-08-15
Limit linted twig directories

bronzehedwick:
Only test the custom folder in modules and themes, to prevent linting contrib.

Re #655


justafish (approved):
Thanks!

justafish (commented):
@bronzehedwick could you provide an example of what this is failing on? It might be better to provide a way to turn it off for a particular file as it's good to know if contrib is providing faulty twig files

justafish (approved):
Thanks!

Issue #650, opened 2024-08-14
drainpipe-dev deploy doesn't work


justafish:
https://github.com/Lullabot/drainpipe/actions/runs/10391050664/job/28773021115

justafish:
https://github.com/Lullabot/drainpipe/actions/runs/10998050354/job/30535081497

YesCT:
This bug came up in #drainpipe  slack: https://lullabot.slack.com/archives/C08MT5PJAF5/p1748990155127909?thread_ts=1748989073.659109&cid=C08MT5PJAF5

mrdavidburns:
@davereid Any updates on this? Do you have availability to work on this?

PR #657, opened 2024-08-16
Issue #650: Fix Drainpipe Dev release workflow

justafish:


justafish:
Will have to test this when we do a release :grimacing:

PR #668, opened 2024-08-23
Replace twig-cs-filter format parameter with report

plopesc:
Running `test:lint` task we are getting the following error:

```
[test:lint]   The "--format" option does not exist.  
[test:lint]                                          
[test:lint] 
[test:lint] lint [-l|--level LEVEL] [-c|--config CONFIG] [-r|--report REPORT] [-f|--fix] [--no-cache] [--debug] [--] [<paths>...]
```
This error is thrown by the `twig-cs-fixer` invocation in line 67.

According to the package [docs](https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/docs/command.md), the parameter used should be `--report` instead of `--format`


justafish:
Fixed in #665

PR #670, opened 2024-08-28
[For dev: update with main] Issue #535: Change multidev deploy to manual only: tag based

YesCT:
### Relates to:
- #535

### Description:
- Run the Multidev Deploy when the label `pantheon-multidev` is added

## Next steps

- Address feedback

## References

From justafish

> Here's the relevant bit of code that will need modifying https://github.com/Lullabot/drainpipe/blob/main/src/ScaffoldInstallerPlugin.php#L285

https://github.com/Lullabot/drainpipe/pull/546#issuecomment-2082805879


scaffold/github/workflows/PantheonReviewAppsManual.yml
```diff
@@ -0,0 +1,72 @@
...
+# You will also need to add the secrets used below to Dependabot secrets
+# if you want this to successfully run in auto-created PRs.
+permissions:
+  deployments: write
+  contents: read
+  pull-requests: read
+
+jobs:
+  Drainpipe-Deploy-Pantheon-Multidev:
+    if: contains(github.event.pull_request.labels.*.name, 'pantheon-multidev')
```
YesCT, line 20:
```suggestion
    if: (contains(github.event.pull_request.labels.*.name, 'pantheon-multidev') && ${{ vars.PANTHEON_MULTIDEV_MANUAL }} == "true") || ${{ vars.PANTHEON_MULTIDEV_MANUAL }} != "true"
```

From justafish

> Could we do something like that instead of duplicating the files? (haven't tested it so not sure if it'll work!)

https://github.com/Lullabot/drainpipe/pull/546#discussion_r1608639828

scaffold/github/workflows/PantheonReviewAppsManual.yml
```diff
@@ -0,0 +1,72 @@
+name: "Pantheon Review Apps"
+
+on:
+  pull_request:
+    types: [ labeled, unlabeled, opened, synchronize, reopened]
```
YesCT, line 5:
```suggestion
    types: [ labeled, opened, synchronize, reopened]
```

From justafish
https://github.com/Lullabot/drainpipe/pull/546#discussion_r1608640047

mrdavidburns:
Is this being used on any projects yet. Feeling good that this is a new Action managed with configuration, but just need to verify the Actions are working as expected. cc @YesCT

mrdavidburns (approved):
@deviantintegral and I reviewed this during our sync yesterday. Since this change is isolated into its own GitHub Action and configurable this would not be a breaking change, just a new feature available to run Pantheon MultiDev manually. Merge approved.

YesCT:
> Is this being used on any projects yet. Feeling good that this is a new Action managed with configuration, but just need to verify the Actions are working as expected. cc @YesCT

Doh. I never answered this question.

A customer ended up testing this as part of the drainpipe release. Ran into an error.
Since then, we have reproduced the error on an internal project (LDC).

I updated the issue summary of https://github.com/Lullabot/drainpipe/issues/535 to indicate we have a PR coming that will fix the manual deploy scaffold action. 🤞

Issue #625, opened 2024-07-30
Drainpipe not reading correctly the URLs for visual diffs


beto-aveiga:
When an override file is used for Tugboat config, it may produce URLs for visual diffs that break Tugboat builds.
If there is a way to build these URLs it should be documented. I tried double quotes and scaping them, but it didn't work.

![image](https://github.com/user-attachments/assets/9c3c959c-43bc-449c-b017-5a674e7d1711)

### On Tugboat

![image](https://github.com/user-attachments/assets/de01389a-edbd-4c71-ac7d-a0b57e671260)

### Tags
composer install, .tugboat/config.drainpipe-override.yml, tugboat, visual diffs, visualdiffs

mrdavidburns:
This would be a blocker now that our static tests are checking for uncommitted changes.

beto-aveiga:
> This would be a blocker now that our static tests are checking for uncommitted changes.
I'm not clear about this @mrdavidburns 

I  just think that overriding URLs with parameters for VIsual Diffs doesn't work as expected.

mrdavidburns:
@beto-aveiga Looks like this is the section of code where query parameters are being encoded. Thoughts on how we can fix that?

https://github.com/Lullabot/drainpipe/pull/441/files#diff-fccaa14362cd8508cb094a15287d2e442fc57d94205c6b2146087a552815d1fcR346-R356
```
            // Filter out unsupported config overrides.
            if (!empty($tugboatConfigOverride['php']) && is_array($tugboatConfigOverride['php'])) {
                $tugboatConfigOverride['php'] = array_filter($tugboatConfigOverride['php'], function($key) {
                    return in_array($key, ['aliases', 'urls', 'visualdiff', 'screenshot']);
                }, ARRAY_FILTER_USE_KEY);
                $overrideOutput = [];
                foreach (explode(PHP_EOL, Yaml::dump($tugboatConfigOverride['php'], 2, 2)) as $line) {
                    $overrideOutput[] = str_repeat(' ', 4) . $line;
                }
                $tugboatConfig['overrides']['php'] = rtrim(implode("\n", $overrideOutput));
            }
```

beto-aveiga:
>  Thoughts on how we can fix that?
My first idea is to encode them before or encode them after, and see what happens when I run `composer install`.
However, Xdebugging and seeing the values in real-time would be easier and more effective for discovering new approaches.

Writing a simple test sounds reasonable, too.

PR #673, opened 2024-08-30
Issue #625: Tugboat urls support query parameters.

mrdavidburns:


src/ScaffoldInstallerPlugin.php
```diff
@@ -407,19 +407,30 @@ private function installTugboat(string $scaffoldPath): void {
...
-                ARRAY_FILTER_USE_KEY);
+            $tugboatConfigOverride['php'] = array_filter($tugboatConfigOverride['php'], function($key) {
+                return in_array($key, ['aliases', 'urls', 'visualdiff', 'screenshot']);
+            }, ARRAY_FILTER_USE_KEY);
+
+            // Prepare YAML data
+            $yamlData = [];
+            foreach ($tugboatConfigOverride['php'] as $key => $value) {
+                $yamlData[$key] = $value;
+            }
```
deviantintegral, line 418: This seems wrong - as in, I suppose it works, but don't you just want something like:

```php
$yamlData = $tugboatConfigOverride['php'];
```

src/ScaffoldInstallerPlugin.php
```diff
@@ -407,19 +407,30 @@ private function installTugboat(string $scaffoldPath): void {
...
         if (!empty($tugboatConfigOverride['php']) && is_array($tugboatConfigOverride['php'])) {
-            $tugboatConfigOverride['php'] = array_filter($tugboatConfigOverride['php'],
-                function($key) {
-                    return in_array($key,
-                        ['aliases', 'urls', 'visualdiff', 'screenshot']);
-                },
-                ARRAY_FILTER_USE_KEY);
+            $tugboatConfigOverride['php'] = array_filter($tugboatConfigOverride['php'], function($key) {
+                return in_array($key, ['aliases', 'urls', 'visualdiff', 'screenshot']);
+            }, ARRAY_FILTER_USE_KEY);
```
deviantintegral, line 412: Just confirming, this appears to only be formatting changes?

deviantintegral (commented):
I'm a little confused by this PR.

At the least, I think doing a broad htmldecode on the YAML itself could lead to other bugs. For example, what happens if you actually want an HTML entity in some other text? What if you are inlining HTML?

Reading the code, it's as if `Yaml::dump()` is encoding HTML entities. But, I did a quick test at https://github.com/deviantintegral/yaml-test and it doesn't:

```
/opt/homebrew/bin/php /Users/andrew/workspace/yaml-test/test.php
urls:
  - '/path?query=value'

Process finished with exit code 0
```

Do you think you can narrow this down a bit into something more targeted that tells us exactly what function is doing the HTML encoding? I like @beto-aveiga 's suggestion of xdebug!

mrdavidburns:
Closing in favor of #681

PR #676, opened 2024-09-02
ISSUE-220: do not use npx for action validator, use binaries

beto-aveiga:
Uses binaries instead of NPX. See #220

### DIFF
--- a/tasks/github.yml
+++ b/tasks/github.yml
@@ -19,13 +19,7 @@ tasks:
           # Fallback for architectures that don't have a release e.g. Windows
           else
               yamls=$(find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \))
-              action_validator_url_base="https://github.com/mpalmer/action-validator/releases/download/v0.6.0/action-validator_"
-              action_validator_bin_path="/usr/local/bin/action-validator"
-              [ "$(arch)" = "x86_64" ] && bin_name="linux_amd64"
-              [ "$(arch)" = "aarch64" ] && bin_name="linux_arm64"
-              curl -Lo $action_validator_bin_path "$action_validator_url_base$bin_name" || true
-              chmod +x $action_validator_bin_path
-              echo $yamls | xargs -n1 $action_validator_bin_path
+              echo $yamls | xargs -n1 action-validator
           fi
 
   composer-lock-diff:
--- a/tasks/github.yml
+++ b/tasks/github.yml
@@ -23,7 +23,7 @@ tasks:
               action_validator_bin_path="/usr/local/bin/action-validator"
               [ "$(arch)" = "x86_64" ] && bin_name="linux_amd64"
               [ "$(arch)" = "aarch64" ] && bin_name="linux_arm64"
-              curl -Lo $action_validator_bin_path "$action_validator_url_base$bin_name") || true'
+              curl -Lo $action_validator_bin_path "$action_validator_url_base$bin_name" || true'
               chmod +x $action_validator_bin_path
               echo $yamls | xargs -n1 $action_validator_bin_path
           fi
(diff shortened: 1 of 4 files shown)


deviantintegral:
I wonder if we should instead install the binary at the same time as all the other binaries we download. It feels a bit odd to be doing this dynamically in a task?

deviantintegral:
Actually, it looks like we already install the binary, so we just have to call vendor/bin/action-validator instead?

beto-aveiga:
@deviantintegral : for the change I made, we are running the `else` part of this line
```
if [ -f ./vendor/bin/action-validator ]; then
```

So, I think there is no action-validator binary at that moment. Am I right?

tasks/github.yml
```diff
@@ -18,7 +18,14 @@ tasks:
...
           else
-              find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \) | xargs -n1 npx @action-validator/cli
+              yamls=$(find $directories -type f \( -iname '*.yaml' -o -iname '*.yml' \))
+              action_validator_url_base="https://github.com/mpalmer/action-validator/releases/download/v0.6.0/action-validator_"
+              action_validator_bin_path="/usr/local/bin/action-validator"
+              [ "$(arch)" = "x86_64" ] && bin_name="linux_amd64"
+              [ "$(arch)" = "aarch64" ] && bin_name="linux_arm64"
+              curl -Lo $action_validator_bin_path "$action_validator_url_base$bin_name" || true
+              chmod +x $action_validator_bin_path
+              echo $yamls | xargs -n1 $action_validator_bin_path
```
deviantintegral, line 28: What do you think about moving the binary install work to https://github.com/Lullabot/drainpipe/blob/main/drainpipe-dev/src/DevBinaryInstallerPlugin.php ? Then it would be available in `vendor/bin/action-validator` like other apps we download.

beto-aveiga:
These two files are very similar.
Do we need a refactor here?

- https://github.com/Lullabot/drainpipe/blob/main/drainpipe-dev/src/DevBinaryInstallerPlugin.php
- https://github.com/Lullabot/drainpipe/blob/9a24bfce962fa3ad24de62e3391033371b2ce192/src/BinaryInstallerPlugin.php#

For instance, if we update the action validator version in one place, it should be the same version in the other binary installer.
We don't need the same binaries, but we need to keep in sync with the versions we are downloading.

beto-aveiga:
@deviantintegral, this is ready for review again. Thanks!

mrdavidburns (approved):
LGTM! Would like us to get a second approval before merging.

deviantintegral:
@beto-aveiga the DevBinaryInstaller only triggers if you're installing drainpipe-dev, which should be in require-dev. Those packages don't get installed when you do composer install --no-dev. So, I think we should move the actions validator binary to the DevBinaryInstaller class.

beto-aveiga:
@deviantintegral, we already have `action-validator` in that class: https://github.com/Lullabot/drainpipe/blob/9a24bfce962fa3ad24de62e3391033371b2ce192/src/BinaryInstallerPlugin.php#L38-L48

beto-aveiga:
@deviantintegral removing action-validator from src/BinaryInstallerPlugin.php is causing an error on the first test.

deviantintegral:
The failure makes sense - the test job isn't installing drainpipe-dev. Let's see if this fixes it.

deviantintegral:
https://github.com/Lullabot/drainpipe/pull/676/commits/e20e3510cf3bd5b2549a5f05ec122e7716aa045f is a great case study in why objects are better than arrays! The actions validator was nested in the wrong section in a way that was not obvious at all.

The tugboat failure looks to be something new, because https://github.com/Lullabot/drainpipe/pull/688 was passing and then I rebuilt it and got the same failure here. _perhaps_ it has to do with something merged recently, but I think it's more likely that there's a problem due to us doing some trickery to move the .tugboat directory around.

PR #679, opened 2024-09-03
Fix upgrading task in task

deviantintegral:
This will fix this error when upgrading task:

```
[drupal:composer:development] 🪠 [Drainpipe] vendor/lullabot/drainpipe/scaffold/env/dotenv.php' missing from autoload-dev files
[drupal:composer:development] 🪠 [Drainpipe] vendor/lullabot/drainpipe/scaffold/env/dotenv.php' missing from autoload-dev files
[drupal:composer:development] Using cached version of task v3.38.0 (30d3c727a434ee3bf69fb69e5d1aa84c3ab401fc2343a2760b4c7808acc689b8)
[drupal:composer:development]
[drupal:composer:development] In Filesystem.php line 372:
[drupal:composer:development]
[drupal:composer:development]   copy(/var/www/html/vendor/bin/task): Failed to open stream: Text file busy
[drupal:composer:development]
[drupal:composer:development]
```

https://github.com/Lullabot/drainpipe/blob/c293ed779f1bcacf64390daa567258f422b582f2/src/BinaryInstaller.php#L253

We need to add a `remove()` call before the copy, so the task binary can be replaced if it is still in use by a process.

Here's a test that should hopefully fail and reproduce this issue: https://github.com/Lullabot/drainpipe/actions/runs/10692591483/job/29641292997

image


.github/workflows/TestBinaryInstaller.yml
```diff
@@ -76,3 +76,12 @@ jobs:
...
           test -x vendor/bin/task
+
+      - name: Prove we can upgrade task within task
+        run: |
+          cd test/fixtures/drainpipe-task-upgrade
+          ddev start
+          ddev composer install
+          ddev task drupal:composer:development
+          cp composer-upgraded.lock composer.lock
+          ddev task drupal:composer:development
```
deviantintegral, line 87: @justafish I wonder if we should consider pulling in bats for testing, which is what ddev uses and is pretty nice for "shell as tests".

.github/workflows/TestBinaryInstaller.yml
```diff
@@ -76,3 +76,12 @@ jobs:
...
           test -x vendor/bin/task
+
+      - name: Prove we can upgrade task within task
+        run: |
+          cd drainpipe/tests/fixtures/drainpipe-task-upgrade
+          ddev start
+          ddev composer install
+          ddev task drupal:composer:development
+          cp composer-upgraded.lock composer.lock
+          ddev task drupal:composer:development
```
deviantintegral, line 87: This isn't working with our fixed code because we're still installing drainpipe from the last release, and not this branch. I need to fix that!

.github/workflows/TestBinaryInstaller.yml
```diff
@@ -76,3 +76,18 @@ jobs:
...
+          cd drainpipe-task-upgrade
+          ddev start
+          ddev composer install
+          ddev task drupal:composer:development
+          cp composer-upgraded.json composer.json
+          cp composer-upgraded.lock composer.lock
+          # Composer won't run updated plugin code until after its been
+          # installed. Once we have two new releases after 3.11.0, that upgrade
+          # task, we can remove this double call.
+          ddev task drupal:composer:development || ddev task drupal:composer:development
```
mrdavidburns, line 93: Clever!

mrdavidburns (approved):

justafish:
nice one :+1:

PR #681, opened 2024-09-04
Output php configurations without escaping them #625

beto-aveiga:
Fixes URLs with parameters for visual diffs in  .tugboat/config.drainpipe-override.yml


deviantintegral (approved):
Nice! 👏

PR #684, opened 2024-09-06
ISSUE-398: fail if there are red lines in the Status Report admin page

beto-aveiga:
Adds a check that fails if there are errors in the Status Report page.

### DIFF
--- a/scaffold/Taskfile.yml
+++ b/scaffold/Taskfile.yml
@@ -42,6 +42,11 @@ tasks:
     cmds:
       # @see https://architecture.lullabot.com/adr/20211026-one-time-login-first-drupal-user-account/
       - ./vendor/bin/drush user:unblock --uid=1
+      - |
+          [ $(drush core:requirements --severity=2 --format=tsv | wc -l) -lt 1 ] \
+          && echo "🟢 Status Report page checked" \
+          || false || echo "🔴 Status Report page check failed"
+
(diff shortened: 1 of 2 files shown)


beto-aveiga:
I tested it using other projects, and it worked.

mrdavidburns (approved):

scaffold/Taskfile.yml
```diff
@@ -42,6 +42,11 @@ tasks:
     cmds:
       # @see https://architecture.lullabot.com/adr/20211026-one-time-login-first-drupal-user-account/
       - ./vendor/bin/drush user:unblock --uid=1
+      - |
+          [ $(drush core:requirements --severity=2 --format=tsv | wc -l) -lt 1 ] \
+          && echo "🟢 Status Report page checked" \
+          || false || echo "🔴 Status Report page check failed"
```
deviantintegral, line 48: What's this `|| false` for?

beto-aveiga: You are right 🤔 
I guess this wasn't needed...
Let me try without it.

deviantintegral (approved):

deviantintegral:
This is probably worth calling out specifically in the release notes since it could cause working builds to fail.

YesCT:
It's listed, but not called out in https://github.com/Lullabot/drainpipe/releases/tag/v3.12.1 cc @mrdavidburns

PR #688, opened 2024-09-11
Fix wrong path to composer.json for TwigExtension cs-fixer

deviantintegral:
This gets copied into the root of the project, and isn't referenced in vendor. This probably needs the web root path updated too.


deviantintegral:
This clearly needs tests because it made it to a release - not only the paths, but that we changed the underlying method calls and didn't notice.

justafish:
#703 for follow up to add tests

PR #704, opened 2024-09-24
Ensure that .drainpipe/gitlab directory is present

plopesc:
Previously, ```````.drainpipe/gitlab directory was created on demand in lines 259 & 269.

However, in local scenarios. line 264 was trying to add the Nightwatch file to the directory before it is created.

To avoid this, and given that Nightwatch.gitlab-ci.yml file is always present, this MR ensures that folder is always created at the beginning of the process.


mrdavidburns (approved):

geekygnr (approved):
This solved an issue I was having with setting up gitlab.

deviantintegral (approved):

deviantintegral:
I think this is failing because it comes from a fork. We should probably clean that up at some point, but for now @plopesc can you refile as a branch in this repo?

mrdavidburns:
Only affects Gitlab. Reviewed and discussed during weekly sync. Merging.

PR #706, opened 2024-09-26
Do not force to scaffold Nightwatch.gitlab-ci.yml

plopesc:
Not all the projects require the Nightwatch.gitlab-ci.yml file.

In fact, in our case, pipelines are broken when this file is included because we get this error in Gitlab.

```
Unable to create pipeline
report-nightwatch job: chosen stage does not exist; available stages are .pre, build, test, deploy, performance, .post
```

Having this file optional and only installed if declared in the composer.json specific section like the rest of gitlab scaffolding files would avoid this error.

Thank you


justafish:
@plopesc thanks! Could you update the documentation to explain how to bring this file in going forward? As it's a breaking change it'll need a major release

plopesc:
@justafish Added a very basic doc regarding the mentioned file. 

I checked #351 to get a better understanding of the needs that new file tries to cover but I was unable to get a better picture.

Let me know if you think that documentation is good enough or needs any improvement.

PR #709, opened 2024-09-27
Test ddev start time with a slim image

deviantintegral:
Testing to see what startup time is like for the CI job based on:

- https://github.com/ddev/ddev/issues/6572
- https://github.com/ddev/ddev/issues/6571
- https://github.com/ddev/ddev/pull/6570
- https://github.com/ddev/ddev/pull/6569


deviantintegral:
The package was set as private, and while I could have figured it out I just made it public for now.

PR #716, opened 2024-09-30
Issue #715: Specify run: once in default scaffold Taskfile.yml

### LINKED ISSUE #715: Use "run: once" by default in the scaffold Taskfile.yml template
We recently got bit by an [unusual behavior in Task](https://github.com/go-task/task/issues/812). When calculating dependencies, Task will re-run the same dependency if it's required by two separate tasks. See the previous link for a full example.

This bit us because we had two separate tasks that each depended on `build:dev`, and one wrapper task that combined the two separate commands into one. The result was that `yarn` was getting run twice, and _in parallel_. The `yarn` command would then sometimes conflict with itself, as one thread modified/deleted files that the other thread expected to be present.

This behavior of running dependent tasks multiple times can be solved a couple of ways, like putting `run: once` within each task that might be executed multiple times.

However, a simpler and less error-prone solution might be to [specify `run: once` in the root of the Taskfile as suggested in that issue](https://github.com/go-task/task/issues/812#issuecomment-1187348509). That gives us a place to document what `run: once` does and why you would want it. Also, if you really had a place where you did want a dependent command to intentionally run multiple times, you can always specify `run: always` on that individual command.

IMO, this should be the default within the scaffold Taskfile.yml. And we should consider putting it at the top of every every individual `task/*.yml` file. Or maybe within each task would be better for that use-case. I would think it would be rare that you would want your SASS to compile twice or run Drupal's update hooks all within the same Task execution. But putting it in the default template seems like a no-brainer that won't impact any existing sites and prevent problems in the future.

quicksketch:
Fixes #715

### DIFF
--- a/scaffold/Taskfile.yml
+++ b/scaffold/Taskfile.yml
@@ -14,6 +14,11 @@ includes:
 output: prefixed
 silent: true
 
+# Only allow tasks to be run once per task invocation. This ensures that
+# multiple dependencies on the same task are only run once.
+# @see https://github.com/go-task/task/issues/812
+run: once
+
 #vars:
 #  DRAINPIPE_SASS: |
 #    web/themes/custom/mytheme/style.scss:web/themes/custom/mytheme/style.css


scaffold/Taskfile.yml
```diff
@@ -14,6 +14,11 @@ includes:
 output: prefixed
 silent: true
 
+# Only allow tasks to be run once per task invocation. This ensures that
+# multiple dependencies on the same task are only run once.
+# @see https://github.com/go-task/task/issues/812
+run: once
```
deviantintegral, line 20: So does this pass down to included task files as well? Or just tasks declared here directly?

quicksketch: I am not sure. The official documentation is at https://taskfile.dev/usage/#limiting-when-tasks-run (maybe we should reference that rather than the GitHub issue). I'm expecting it only works within the current file, considering [variables need to be manually passed to includes](https://taskfile.dev/usage/#vars-of-included-taskfiles).

deviantintegral (approved):

Issue #370, opened 2024-01-30
Add validation and warnings for composer-patches configuration


deviantintegral:
While we probably shouldn't enforce these ADRs, we could certainly log a warning if any of these configurations are not as expected.

I think we could do this in a composer install hook?

https://architecture.lullabot.com/adr/20220429-composer-patchlevel/
https://architecture.lullabot.com/adr/20220429-composer-patches-inline/
https://architecture.lullabot.com/adr/20220429-composer-patch-files/
https://architecture.lullabot.com/adr/20220429-composer-exit-failure/

beto-aveiga:
@deviantintegral: regarding https://architecture.lullabot.com/adr/20220429-composer-patches-inline/ ... and considering we have a tool like Drainpipe, why not allow the user to use a different patch file and join the composer file with the patches file at the moment of validation?
`composer.json` already has enough information, and if we can have a separation for patches and have a full validation of the composer contents, that would be the best scenario for developers.

mrdavidburns:
I'd prefer if this lives in the composer-patches project. I've created an issue for this.
* https://github.com/cweagans/composer-patches/issues/617

mrdavidburns:
We also have a repo set up if this needed to be separate from composer-patches.

https://github.com/Lullabot/drainpipe/pull/717#issuecomment-2498153474
https://github.com/Lullabot/composer-checks

mrdavidburns:
Once [https%3A//github.com/Lullabot/composer%2Dchecks ](https://github.com/Lullabot/composer-checks) is ready and pushed to packagist we can come back to integrate this w/ Drainpipe.

beto-aveiga:
OK, it seems I have nothing to do here 👀

deviantintegral:
@beto-aveiga there's three open issues at https://github.com/Lullabot/composer-checks/issues that we need done I think.

PR #717, opened 2024-10-01
ISSUE-370: composer configuration checks

beto-aveiga:
### WORK FOR THIS IS BEING DONE IN https://github.com/Lullabot/composer-checks/issues

relates to: #370

It adds the following checks when running `composer install`.

https://architecture.lullabot.com/adr/20220429-composer-patchlevel/
https://architecture.lullabot.com/adr/20220429-composer-patches-inline/
https://architecture.lullabot.com/adr/20220429-composer-patch-files/
https://architecture.lullabot.com/adr/20220429-composer-exit-failure/

It also adds a test for the checks.

### Screenshots

image


.github/workflows/TestScaffoldInstallerPlugin.yml
```diff
@@ -0,0 +1,108 @@
...
+          ddev composer config --unset extra.patchLevel
+          check_composer_install_contents "Configure Composer patches to use \`-p2\` as \`patchLevel\` for Drupal core" 1
+
+          ddev composer config extra.patches-file "composer.patches.json"
+          check_composer_install_contents "Store Composer patches configuration in \`composer.json\`" 1
+
+          ddev composer config --unset extra.patches-file
+          check_composer_install_contents "Store Composer patches configuration in \`composer.json\`" 0
+
+          ddev composer config extra.patches --json '{"drupal":{"issue-x":"http"}}'
```
mrdavidburns, line 104: This is a great addition that aligns with our [ADR: Use local copies of patch files](https://architecture.lullabot.com/adr/20220429-composer-patch-files/)

I know we have completely updated all our projects to adhere to this, so having a check that will enforce this would be a nice addition.

mrdavidburns (approved):
I approve this feature!

justafish (changes requested):
Very cool!

Could we have this on by default and allow an opt-out?

src/ScaffoldInstallerPlugin.php
```diff
@@ -72,6 +72,130 @@ public static function getSubscribedEvents()
...
 
+    /**
+     * Composer configuration advice: Use local copies of patch files.
+     *
+     * @return void
+     */
+    private function _checkComposerPatchesAreLocal()
+    {
+        $patchesInComposer = $this->extra['patches'] ?? false;
+        $patchesInExtraFile = $this->extra['patches-file'] ?? false;
```
justafish, line 83: I think we should throw up a warning and discourage this as well since it won't get picked up by `composer validate` when patches are in a separate file

src/ScaffoldInstallerPlugin.php
```diff
@@ -72,6 +72,130 @@ public static function getSubscribedEvents()
         ];
     }
 
+    /**
+     * Composer configuration advice: Use local copies of patch files.
+     *
+     * @return void
+     */
+    private function _checkComposerPatchesAreLocal()
```
deviantintegral, line 80: I don't think we need to prefix these with an underscore - they're private methods?

src/ScaffoldInstallerPlugin.php
```diff
@@ -72,6 +72,130 @@ public static function getSubscribedEvents()
...
+
+        // Patches defined on a separate file.
+        if ($patchesInExtraFile) {
+            if (!file_exists($patchesInExtraFile)) {
+                $this->io->warning("The patches file `$patchesInExtraFile` can't be read.");
+                return;
+            }
+
+            $patchesJsonEncoded = file_get_contents($patchesInExtraFile);
+            $patchesContent = json_decode($patchesJsonEncoded, true)['patches'] ?? [];
```
deviantintegral, line 93: I think we need to either check `json_last_error()` or use an existing decoder utility function.

src/ScaffoldInstallerPlugin.php
```diff
@@ -72,6 +72,130 @@ public static function getSubscribedEvents()
...
+            !$composerExitsOnPatchFailureBool;
+        $condition2 = $composerExitsOnPatchFailureBool
+            && $composerExitsOnPatchFailure !== true;
+        $warn = $condition1 && $condition2;
+
+        if (!$warn) {
+            return;
+        }
+
+        $msg = 'Break Composer install if patches don\'t apply. See';
```
deviantintegral, line 157: Make this `"` so you don't have to escape?

src/ScaffoldInstallerPlugin.php
```diff
@@ -72,6 +72,130 @@ public static function getSubscribedEvents()
...
+            return;
+        }
+
+        $msg = 'Configure Composer patches to use `-p2` as `patchLevel` for Drupal core. See';
+        $link = 'https://architecture.lullabot.com/adr/20220429-composer-patchlevel/';
+        $this->io->warning("$msg $link");
+    }
+
+    /**
+     *  Composer configuration advice: "patches-file" is not set/used.
```
deviantintegral, line 184:
```suggestion
     * Composer configuration advice: "patches-file" is not set/used.
```

deviantintegral (commented):
This is great! I love codifying our ADRs like this 👏 .

I wonder if it's worth moving this to a new plugin class. The other code in this class is installing scaffold files, while this is validating composer which feels like a different concern.

src/ScaffoldInstallerPlugin.php
beto-aveiga: Using a patches-file is discouraged on a separate check on this PR.

src/ScaffoldInstallerPlugin.php
beto-aveiga: I got a VSCode hint when I started moving the code to the Drainpipe repo branch, and I thought it was a coding standard in the Drainpipe project. It is probably just a misconfiguration on my IDE. I will remove the `_` from the checks I created.

beto-aveiga:
> @justafish  Could we have this on by default and allow an opt-out?

Opt-out using composer configuration in the "extra" section?  Do you mean something like this?

image

src/ScaffoldInstallerPlugin.php
beto-aveiga: Will do. Thanks!

beto-aveiga:
> @deviantintegral I wonder if it's worth moving this to a new plugin class. The other code in this class is installing scaffold files, while this is validating composer which feels like a different concern.

I think the answer is yes. Definitely.
I will create a  `ComposerChecksPlugin` class.

beto-aveiga:
If tests pass, the only thing missing is the opt-out configuration. https://github.com/Lullabot/drainpipe/pull/717#pullrequestreview-2345451265

beto-aveiga:
I also added documentation for the opt-out configurations.

beto-aveiga:
Added tests for the opt-out options.

justafish:
@beto-aveiga - @deviantintegral and I discussed this and we think it'd be really cool to have as a separate project as it doesn't necessarily need Drupal. We can then pull it into Drainpipe as a requirement. We've created a repo for you here and invited you: https://github.com/Lullabot/composer-checks

Thanks!

.github/workflows/TestComposerChecksPlugin.yml
```diff
@@ -0,0 +1,60 @@
+name: ComposerChecksPlugin - Checks Composer checks are okay
```
mrdavidburns, line 1: @deviantintegral Did we want this to be its own Action or would it make sense to bundle it in w/ `TestStatic`???

mrdavidburns: Leaving this separate would add ~2m45s each time it runs.

deviantintegral: I agree, let's bundle it in with TestStatic.

README.md
```diff
@@ -645,6 +645,63 @@ php:
   visualdiff:
 ```
 
+## Composer configuration checks
+
+To promote best practices within Drupal/Composer projects, we run a few checks after running Composer commmads.
```
deviantintegral, line 650: Let's move all of these docs to the composer-checks readme.

composer.json
```diff
@@ -3,6 +3,12 @@
     "description": "An automated build tool to allow projects to have a set standardized operations scripts.",
     "license": "GPL-2.0-or-later",
     "type": "composer-plugin",
+    "repositories": [
+        {
+            "type": "git",
+            "url": "https://github.com/Lullabot/composer-checks.git"
+        }
+    ],
```
deviantintegral, line 11: As you see from the above CI checks, `repositories` is a root-only value. Instead, let's publish the plugin to packagist like a normal plugin.

https://github.com/Lullabot/composer-checks/issues/3

composer.json
```diff
@@ -16,7 +22,8 @@
         "drush/drush": "^11|^12|^13",
         "symfony/yaml": "^6|^7",
         "twig/twig": "^3",
-        "vlucas/phpdotenv": "^4|^5"
+        "vlucas/phpdotenv": "^4|^5",
+        "lullabot/composer-checks": "@dev"
```
deviantintegral, line 26: Let's create a tag we can point to instead.

https://github.com/Lullabot/composer-checks/issues/4

.github/workflows/TestComposerChecksPlugin.yml
deviantintegral: Actually, I think this is for testing the plugin code itself. TestStatic (and everything else) will trigger any errors automatically as soon as they run composer install.

.github/workflows/scripts/ComposerChecks.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+check_composer_install_contents() {
+  local string_to_check="$1"
+  local expected="$2"
+  local file="composer-install-output.txt"
+
+  ddev composer install > $file 2>&1
```
deviantintegral, line 8: Let's move all of these tests to the composer-checks repository. They don't need to be drainpipe specific. In fact, it would be reasonable to simplify this and use github's setup-php action to test this without ddev.

Some? All? of this may be better as a phpunit test too. Here's an example: https://github.com/deviantintegral/composer-gavel/blob/main/tests/ComposerVersionRequirementTest.php

deviantintegral (commented):
This is getting there! I filed issues in the other project too: https://github.com/Lullabot/composer-checks/issues

deviantintegral:
@beto-aveiga The composer checks package is now published to packagist. Let's pull that into Drainpipe as a new dev dependency in a new PR. Thanks!

PR #740, opened 2024-11-25
Justafish/issue 204 acquia integration  v3

beto-aveiga:
Adds a Deploy to Acquia Github workflow (with documentation).


beto-aveiga:
The tests are passing. This was the first step to accomplish. 😌 

Also, the DEV environment is working as expected. 🎉 
http://lullabotsandboxffi2ugpgwh.devcloud.acquia-sites.com/

I found a few things difficult, to list a few of them:
- Knowing what is being executed in what context is a bit complex when running Github Workflows, more so when you run actions inside actions. I wonder if we could, whenever we run an action or source a script, print the hierarchy of files being called to understand better what is happening.
- Understand that deploying is beyond code, it is also about state, so to test an Acquia Deployment, there are conditions in both layers (code and state) that must be met. Sometimes, I needed code changes; sometimes, I needed DB changes.
- Because the deployment copies DB from staging, I needed to start with a working staging environment because we are cloning from it. Starting with a working staging environment required to mimic the Drupal install locally, then pushing to the Acquia repo, then changing the code in staging, then running `drush deploy`; basically, a manual deployment.
- When downloading drush aliases, a path for drush is set; sometimes, it is a path to drush9. Drush9 won't always work, so you must remove/update those paths to avoid using older drush versions. This seems to be an Acquia issue: https://docs.acquia.com/acquia-cloud-platform/manage-apps/command-line/drush#section-specifying-a-version-of-drush-in-commands

### What is next?
Try the deployment with a real site.

### Then... ?
Send this PR for review. I guess that before sending this PR for review, I will need to update both PRs with current Drainpipe main's branch code to ease spotting the differences between this PR and 

@mrdavidburns @quicksketch @justafish

.github/workflows/TestAcquia.yml
```diff
@@ -70,10 +90,10 @@ jobs:
           echo "/drainpipe" >> .drainpipeignore
           ddev task snapshot:directory directory=/tmp/release
 
-      - name: Clone from production to dev
+      - name: Clone from test to dev
         uses: ./drainpipe/scaffold/github/actions/acquia/clone-env
         with:
-          source-environment: lullabotsandbox.prod
+          source-environment: lullabotsandbox.test
```
mrdavidburns, line 96: @beto-aveiga were we able to get this working w/ prod after we notified we were having issues testing with prod in our Sandbox?

beto-aveiga: Sure! I can try it now.
I think there will be no issues.

beto-aveiga:
The tests are passing, but yesterday, I noticed that the configuration is not being imported.
I wonder if this is related to code-switching needing a container replacement when the code is switched, and we are running configuration imports too early.
By the way, I'm cloning from PROD to DEV. PROD works as expected.

@mrdavidburns

beto-aveiga:
> @beto-aveiga I notice you have a lot of Github Action workflows deleted in this pull request. Was that to save time while testing?

Absolutely. I will restore them when there is no need for more code.

> What's the intent of this pull request? Will we cherry-pick the important things to keep like:

Make `.github/workflows/TestAcquia.yml` to pass and ensure the deployment works in the DEV environment where we arr deploying. I'm preserving the original test structure that @justafish wrote.


@mrdavidburns

beto-aveiga:
At `scaffold/github/actions/acquia/clone-env/action.yml`, we run:

```
        drainpipe_exec "acli env:mirror --no-config --no-interaction ${{ inputs.source-environment }} ${{ inputs.target-environment }}"
```

This is likely an async command, so it will run after it returns the control to our deployment script. If we immediately run configuration imports, those configuration imports will be lost because our application's container(s) will be replaced. I will check if this is true by looking at the deploy workflow and the Acquia activity feed.


@mrdavidburns

beto-aveiga:
Regarding my previous comment...
I re-ran the deployment process and verified that after our deploy workflow test had ended, the Acquia code switch was still running. That means that the code switch was still in progress while we were running the configuration updates.

In other words, after we ran acquia/push, we ran acquia/update, but acquia/push invoked `acli api:environments:code-switch`, which happens asynchronously and takes a few minutes.

So, between acquia/push and acquia/update we need an something to way for the code switch to complete, otherwise our config import changes will be lost.

mrdavidburns:
@beto-aveiga Would this work?
https://docs.acquia.com/acquia-cloud-platform/add-ons/acquia-cli/commands/app:task-wait

beto-aveiga:
I love that solution. Thanks. @mrdavidburns !

I was already asking ChatGPT, and it gave me a different route.
https://chatgpt.com/share/67a0f0b2-fe3c-8004-b414-df645e3d24e0

But I will first try yours because it looks self-explanatory.

beto-aveiga:
Github status says everything is okay but workflows are not running 🤔

beto-aveiga:
Good! 
So far, it's working as expected. :) 
image

beto-aveiga:
Taking forever... I checked Acquia UI, also taking forever.

beto-aveiga:
Re-running it. I don't know why the same actions took forever to complete. Then only thing I added was `--task-wait`

beto-aveiga:
Now that it works, the next step is to try to make a deployment to DEV with an actual site repository on Acquia.
Then, if that works as expected, finish this PR by restoring files and cleaning the code.

mrdavidburns:
Let's add a section to the README that explains the Acquia integration is Beta but should not break any other functionality that is already in place.

scaffold/github/actions/common/set-env/action.yml
```diff
@@ -12,6 +12,15 @@ runs:
   steps:
     - run: |
         echo "Setting Drainpipe environment variables:"
-        echo "DRAINPIPE_PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')" >> $GITHUB_ENV
-        echo "DRAINPIPE_SHA=$(cat $GITHUB_EVENT_PATH | jq -r .pull_request.head.sha)" >> $GITHUB_ENV
+
```
beto-aveiga, line 15: I will try to revert this one after/if all tests pass now.

beto-aveiga: Reverted. I will try a deploy to DEV on a real site to see if deploying still works

beto-aveiga: The deployment didn´t work when I restored the file.
The changes in this PR are needed.

PR #745, opened 2024-12-03
fix(#743): "drupal:db-import" frozen if $DB_DIR/db.sql.gz already exists

### LINKED ISSUE #743: Never completing task drupal:import-db
Problem:

I've run `ddev drupal:import-db`, forgot about something, pressed Cntrl+c.

Fix my stuff,  `ddev drupal:import-db` nothing happens. My dump is huge, so didn't notice until a while after (as I already expected it would take a while), pressed Cntrl+c and THEN I see:

```
[drupal:import-db] 📰 Importing database
^Ctask: Signal received: "interrupt"
[drupal:import-db] gzip: /var/www/html/files/db/db.sql already exists; do you wish to overwrite (y or n)?
```

Checked `drupal.yml`:

```
      - gunzip -k $DB_DIR/db.sql.gz
```

Checked gunzip options:

```
gunzip
  -f, --force       force overwrite of output file and compress links
  -k, --keep        keep (don't delete) input files
```

Should we use -kf here?

penyaskito:
Fixes #743.

### DIFF
--- a/tasks/drupal.yml
+++ b/tasks/drupal.yml
@@ -44,7 +44,7 @@ tasks:
       - echo "🚮 Dropping existing database"
       - ./vendor/bin/drush {{ .site }} sql:drop --yes
       - echo "📰 Importing database"
-      - gunzip -k $DB_DIR/db.sql.gz
+      - gunzip --keep --force $DB_DIR/db.sql.gz
       - ./vendor/bin/drush {{ .site }} sql:query --file=$DB_DIR/db.sql
       - defer: rm -f $DB_DIR/db.sql
   export-db:


justafish:
Thanks!

tasks/drupal.yml
```diff
@@ -44,7 +44,7 @@ tasks:
       - echo "🚮 Dropping existing database"
       - ./vendor/bin/drush {{ .site }} sql:drop --yes
       - echo "📰 Importing database"
-      - gunzip -k $DB_DIR/db.sql.gz
+      - gunzip --keep --force $DB_DIR/db.sql.gz
```
deviantintegral, line 47: Seems reasonable, and good fix using the long flags!

PR #747, opened 2024-12-04
Allow private keys to be provided to GitLab in base64 format

justafish:


scaffold/gitlab/DDEV.gitlab-ci.yml
```diff
@@ -28,7 +28,10 @@ cache:
     - |
       mkdir -p .ddev/homeadditions/.ssh
       # Copy private key
-      if [ "$DRAINPIPE_DDEV_SSH_PRIVATE_KEY" != "" ]; then
+      if [ "$DRAINPIPE_SSH_PRIVATE_KEY_BASE64" != "" ]; then
+        echo "$DRAINPIPE_SSH_PRIVATE_KEY_BASE64" | base64 -d > .ddev/homeadditions/.ssh/id_rsa
```
deviantintegral, line 32: Is this really an rsa key? Or should it be an ed25519 key?

justafish: Opened a follow up issue here https://github.com/Lullabot/drainpipe/issues/766 as I think this will need changing in a few places, plus allow you to select either option

PR #758, opened 2024-12-16
Exclude include fixer from twigcsfixer

mark-casias:
Removes the Include function from the TwigCSFixer which causes issues in Storybook.


drainpipe-dev/scaffold/.twig-cs-fixer.dist.php
```diff
@@ -32,6 +32,9 @@
 // default standard.
 $ruleset->addStandard(new TwigCsFixer\Standard\TwigCsFixer());
 
+// Remove the IncludeFunctionRule which breaks TwigJS.
+$ruleset->removeRule(TwigCsFixer\Rules\Function\IncludeFunctionRule::class);
```
deviantintegral, line 36: Is it possible to do this by committing `.twig-cs-fixer.php` to your local repo? Then you could control this and anything else (since this twigjs isn't commonly used).

mark-casias: Agreed. This would be a good thing to just document over changing drainpipe... that is, assuming drainpipe doesn't overwrite files that are present, or the scaffolding can be controlled in composer.

mrdavidburns:
@mark-casias Would #756 still need this patch?

mark-casias:
I don't believe so. If we commit the twig-cs-fixer.php file as @deviantintegral suggests, we can close this. Actually since the PR is available in the lullabot-d8 repo, we can close this anyway..

mrdavidburns:
Closing based on @mark-casias last comment.

Issue #757, opened 2024-12-16
Nightwatch Test Failures When Configuring Redis with Drainpipe


rene-ipiales-lullabot:
When configuring Redis with Drainpipe, Nightwatch tests fail due to conflicts in the cache configuration. This issue occurs because the default cache prefix is not explicitly set for each browser, leading to potential cache overlap and test instability.

### Proposed Solution:

* To resolve this issue, add the following configuration to the respective browser-specific settings files:
- For Chrome `/scaffold/nightwatch/chrome.settings.php`
$settings['cache_prefix']['default'] = 'chrome-redis';

- For Firefox `/scaffold/nightwatch/firefox.settings.php`
$settings['cache_prefix']['default'] = 'firefox-redis';

### Outcome:

This ensures each browser has a distinct cache prefix, preventing conflicts and stabilizing Nightwatch tests.

#### Additional Context:

Issue observed in the Drainpipe setup during Redis configuration.
Nightwatch test failures occur consistently without the proposed fix.

PR #779, opened 2025-02-14
Issue #757: Add Cache prefix on the chrome/firefox settings files.

alexanderpatriciop:
### Relates to:
- #757 

### Description:
- Add Cache prefix on the chrome/firefox settings files.


mrdavidburns (approved):

mrdavidburns (approved):
Avoid cache pollution between testing, looks good to me.

Issue #677, opened 2024-09-03
Avoid composer lock diff action if composer.lock hasn't changed


beto-aveiga:
Composer lock diff action takes about ~3 minutes on a minimal site.
There is no need to run this workflow if there are no composer changes, or we could avoid running it for 3 minutes (depending on the site, this may vary).
We could skip it in the mentioned scenario and use less than 1 minute.

deviantintegral:
Good call. I think we can do this with a path filter in GitHub at least: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

alexanderpatriciop:
@deviantintegral Good suggestion, I tested it on a test project and it works as desired.

PR #780, opened 2025-02-17
Issue #677: Avoid composer lock diff action if composer.lock has not changed.

alexanderpatriciop:
### Relates to:
- #677 

### Description:
- Avoid composer lock diff action if composer.lock has not changed.

### DIFF
--- a/scaffold/github/workflows/ComposerLockDiff.yml
+++ b/scaffold/github/workflows/ComposerLockDiff.yml
@@ -3,6 +3,8 @@ name: "Composer Lock Diff"
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+    paths:
+      - 'composer.lock'
 
 concurrency:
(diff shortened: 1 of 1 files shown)


scaffold/github/workflows/ComposerLockDiff.yml
```diff
@@ -3,6 +3,8 @@ name: "Composer Lock Diff"
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+    paths:
+      - 'composer.lock'
```
mrdavidburns, line 7: I had to look this up. This is a really smart simple solution to only execute this workflow if the file had changed. Nice!

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

mrdavidburns (approved):

Issue #797, opened 2025-03-20
Investigate Renovate Config for packages that we support a Range


mrdavidburns:
Drainpipe currently supports Drush 11 - 13, but Renovate had a pull request that tried to changes this to only support v13.
Let's see if there's a way we can improve `renovate.json` so that it recognizes ranges. This way we don't have a backlog of pending updates on the Dependency Dashboard.

![Image](https://github.com/user-attachments/assets/4f6e177f-e049-41ce-a9e4-02fbeb05793c)

mrdavidburns:
From Google AI Studio

Yes, Mend Renovate supports ranges for package versions, and it understands the syntax you provided (`drush/drush:^11|^12|^13`).  Here's a breakdown of how it works and some important nuances:

**How Renovate Interprets Ranges**

Renovate uses the `npm`'s `semver` library (or an equivalent for other package managers) to parse and resolve version ranges.  This means it adheres to standard Semantic Versioning (SemVer) conventions.  The key operators in your example are:

*   **`^` (Caret):**  The caret allows compatible updates within a major version.
    *   `^11`:  This means any version from `11.0.0` up to, *but not including*, `12.0.0`.  It will allow updates to `11.1.0`, `11.2.5`, `11.99.99`, etc.
    *   `^0.2`:  This is a special case.  Before `1.0.0`, semver treats minor version bumps as *potentially* breaking.  `^0.2` would allow updates to `0.2.1`, `0.2.2`, but *not* `0.3.0`.
    *   `^0.0.3`:  Before reaching `0.1.0`, even patch releases are considered potentially breaking. `^0.0.3` will allow updates to versions >= `0.0.3` and < `0.0.4`.

*   **`|` (Pipe):** The pipe acts as an "OR" operator.  It allows you to specify multiple, distinct ranges.  `^11|^12|^13` is saying "match any version that satisfies `^11` OR `^12` OR `^13`".

**Example: `drush/drush:^11|^12|^13`**

This specific range will allow Renovate to create pull requests for any Drush version that falls within one of these ranges:

*   `>=11.0.0 <12.0.0`
*   `>=12.0.0 <13.0.0`
*   `>=13.0.0 <14.0.0`

If the current version is `11.5.0`, and `11.6.0`, `12.7.0`, and `13.1.0` are released, Renovate would (by default) propose an update to the highest compatible version, following these steps:

1.  **Determine valid versions:** Renovate identifies all available versions.
[...]

PR #798, opened 2025-03-21
Issue #797: improving renovate config to support ranges and ignoring drainpipe releases.

mrdavidburns:
Resolves: #797 

This pull request includes updates to the `renovate.json` file to manage package dependencies more effectively. The most important changes involve adding specific rules for certain packages and setting version constraints.

Dependency management updates:

* Added a rule to disable updates for `lullabot/drainpipe` and `lullabot/drainpipe-dev` packages.
* Added a rule to allow versions `^11`, `^12`, and `^13` for the `drush/drush` package.
* Added a rule to allow versions `^6` and `^7` for packages matching the pattern `^symfony/`.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -20,6 +20,21 @@
       "addLabels": [
         "test only - do not merge"
       ]
+    },
+    {
+      "matchPackageNames": [
+        "lullabot/drainpipe",
+        "lullabot/drainpipe-dev"
+      ],
+      "enabled": false
+    },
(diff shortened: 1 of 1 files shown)


deviantintegral (approved):
This looks good. I wish JSON let you have comments so we could cross-reference the composer.json file, but 🤷 .

Let’s see what happens when Renovate runs on main after merging!

Issue #771, opened 2025-02-03
Add static analysis for Github Action files


mrdavidburns:
Add security checks for Github Actions as action-validator is already integrated.

https://github.com/woodruffw/zizmor will add additional security checks. Let's add it in to the `github:actions:lint` task.

elvism-lullabot:
Is it necessary to integrate the [Zizmor](https://github.com/woodruffw/zizmor) plugin into drainpipe in the same way as was done for @action-validator?

mrdavidburns:
We discussed during our sync. This is already being done, as @elvism-lullabot has explained above.

deviantintegral:
While actions-validator is there, let's add zizmor to do additional security checks on actions.

alexis-saransig-lullabot:
I've tried integrating Zizmor in several different ways so far, but none have fully worked within GitHub Actions:

- Using uv to run Zizmor via Cargo

- Using the official Zizmor Docker image

I mount the repository and run the container directly `docker run --rm -v $(pwd):/repo ghcr.io/zizmorcore/zizmor:1.9.0 /repo/.github/workflows /repo/.github/actions`, this works fine locally and generates the expected results, but running it within a GitHub Action doesn't show any visible results. It works fine when run manually or locally, but I haven't been able to get consistent results using the GitHub Action.

![Image](https://github.com/user-attachments/assets/889556bd-6a55-44b6-97ea-a83fb9c95aae)

This likely requires further investigation, and I'm open to suggestions. 
cc: @mrdavidburns

alexis-saransig-lullabot:
> [@alexis-saransig-lullabot](https://github.com/alexis-saransig-lullabot) Have you looked into this https://github.com/fabasoad/setup-zizmor-action?

No, I haven't. I'll try it when I'm back (finishing other project tickets). Thanks David.

elvism-lullabot:
I'll continue with another attempt using @mrdavidburns  suggestions.

elvism-lullabot:
@mrdavidburns @deviantintegral You can find the list of code scanning results from the Zizmor action for this repository [here](https://github.com/Lullabot/drainpipe/security/code-scanning?page=1&query=pr%3A799+tool%3Azizmor+is%3Aopen). Most of them can be remediated following the recommendations [here](https://docs.zizmor.sh/audits/#template-injection)

mrdavidburns:
[...]

PR #799, opened 2025-03-25
Issue #771: Add static analysis with zizmor.

elvism-lullabot:
See #771

### DIFF
--- a/scaffold/github/actions/common/zizmor/action.yml
+++ b/scaffold/github/actions/common/zizmor/action.yml
@@ -1,0 +1,33 @@
+name: Security Analysis with zizmor
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["*"]
+
+permissions: {}
+
+jobs:
+  zizmor:
+    name: zizmor latest via Cargo
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      security-events: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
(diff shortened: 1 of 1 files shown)


scaffold/github/workflows/TestZizmor.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["*"]
+
+jobs:
+  zizmor:
+    name: Run Zizmor Security Analysis
+    runs-on: ubuntu-latest
```
deviantintegral, line 12: Pin the ubuntu version, so 26 doesn't break things later?

```suggestion
    runs-on: ubuntu-24.04
```

mrdavidburns:
@elvism-lullabot Let's update this pull request so that it only scaffolds this file when Drainpipe is configured to run security checks. 

see: https://github.com/Lullabot/drainpipe?tab=readme-ov-file#security

Also, we need to make sure all checks are passing before this can be merged.
image

mrdavidburns:
I'm still not 100% this should be shipped w/ Drainpipe. If we do it would need to be configurable and something we can turn on with configuration.

deviantintegral:
@elvism-lullabot can you file a separate PR just to do the bit around pinning remote workflows? Like these lines:

`webfactory/ssh-agent@a6f90b1f127823b31d4d4a8d96047790581349bd #v0.9.1`

rabbitlair:
Thanks for the review! I will be merging this one, then I will open a different pull request to scaffold Zizmor files as part of Drainpipe (when configured), and to fix any findings on GitHub Actions provided by Drainpipe

Issue #752, opened 2024-12-10
Redis for Tugboat generates wrong image file name


rene-ipiales-lullabot:
Version https://github.com/Lullabot/drainpipe/releases/tag/v3.12.0

## Steps to reproduce
* Run `ddev get ddev/ddev-redis` 
* We will get this file `docker-compose.redis.yaml` 

![image](https://github.com/user-attachments/assets/d6abcd4c-0d11-45eb-9477-bdc4b75197d3)
* Run `composer install` 

* The `.tugboat/config.yml` will be updated with this `Redis` image name 

![image](https://github.com/user-attachments/assets/3ce02445-aee4-44f7-abdc-bc5af698bc8b)

PR #837, opened 2025-05-07
Issue #752: Fix redis image name for tugboat.

rene-ipiales-lullabot:
resolves: https://github.com/Lullabot/drainpipe/issues/752

## Problem

The logic to extract the Redis image version from `.ddev/docker-compose.redis.yaml` only supported one format, resulting in incorrect or unknown versions being assigned to the configuration in some setups.

## Background

DDEV has changed the way Redis images are defined in different versions:

- **In [`v1.2.3`](https://github.com/ddev/ddev-redis/blob/v1.2.3/docker-compose.redis.yaml)**, the image is defined as:

  ```yaml
  image: redis:${REDIS_TAG:-6-bullseye}
  ```

- **In [v2.0.2](https://github.com/ddev/ddev-redis/blob/v2.0.2/docker-compose.redis.yaml)**, the image is defined as:

  ```yaml
  image: ${REDIS_DOCKER_IMAGE:-redis:7}
  ```

Updated the image parsing logic to support both patterns:

* `${REDIS_DOCKER_IMAGE:-redis:7}`

* `redis:${REDIS_TAG:-6-bullseye}`

It now correctly resolves the fallback tag regardless of which format is present.  Additionally, fixes the wrong tugboat name

image

### DIFF
--- a/src/ScaffoldInstallerPlugin.php
+++ b/src/ScaffoldInstallerPlugin.php
@@ -468,10 +468,26 @@ EOT;
         // Add Redis service.
         if (file_exists('./.ddev/docker-compose.redis.yaml')) {
             $redisConfig = Yaml::parseFile('.ddev/docker-compose.redis.yaml');
-            $image = $redisConfig['services']['redis']['image'] ?? '';
-            preg_match('/:(\d+)/', $image, $matches);
-            $tugboatConfig['memory_cache_version'] = $matches[1] ?? 'unknown';
+            $imageRaw = $redisConfig['services']['redis']['image'] ?? '';
+
+            // Normalize image from possible environment syntax.
+            if (preg_match('/^\$\{[^:}]+:-(.+)\}$/', $imageRaw, $matches)) {
+                $image = $matches[1];
+            } elseif (preg_match('/^([^:]+):\$\{[^:}]+:-(.+)\}$/', $imageRaw, $matches)) {
+                $image = "{$matches[1]}:{$matches[2]}";
+            } else {
+                $image = $imageRaw;
+            }
+
+            // Extract the version/tag from the image string.
+            if (preg_match('/:([a-zA-Z0-9._-]+)$/', $image, $versionMatch)) {
+                $version = $versionMatch[1];
+            } else {
+                $version = 'unknown';
+            }
+
             $tugboatConfig['memory_cache_type'] = 'redis';
+            $tugboatConfig['memory_cache_version'] = $version;
         }
 
         // Add Elasticsearch service.
--- a/src/ScaffoldInstallerPlugin.php
+++ b/src/ScaffoldInstallerPlugin.php
@@ -468,26 +468,26 @@ EOT;
         // Add Redis service.
         if (file_exists('./.ddev/docker-compose.redis.yaml')) {
             $redisConfig = Yaml::parseFile('.ddev/docker-compose.redis.yaml');
-            $imageRaw = $redisConfig['services']['redis']['image'] ?? '';
-
+            $image = $redisConfig['services']['redis']['image'] ?? '';
+        
             // Normalize image from possible environment syntax.
-            if (preg_match('/^\$\{[^:}]+:-(.+)\}$/', $imageRaw, $matches)) {
+            // Example: $image = '${REDIS_DOCKER_IMAGE:-redis:7}'.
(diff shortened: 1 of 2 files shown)


src/ScaffoldInstallerPlugin.php
```diff
@@ -468,10 +468,26 @@ function($key) {
...
+                $image = "{$matches[1]}:{$matches[2]}";
+            } else {
+                $image = $imageRaw;
+            }
+
+            // Extract the version/tag from the image string.
+            if (preg_match('/:([a-zA-Z0-9._-]+)$/', $image, $versionMatch)) {
+                $version = $versionMatch[1];
+            } else {
+                $version = 'unknown';
```
mrdavidburns, line 486: @deviantintegral Should we return unknown or have a fallback Redis version?

deviantintegral: Hm, I think this is tricky because if we hit this case we _know_ ddev start will fail. Why don't we throw a `\RuntimeException` instead, since we know a developer will have to intervene?

src/ScaffoldInstallerPlugin.php
```diff
@@ -468,10 +468,26 @@ function($key) {
         // Add Redis service.
         if (file_exists('./.ddev/docker-compose.redis.yaml')) {
             $redisConfig = Yaml::parseFile('.ddev/docker-compose.redis.yaml');
-            $redisImage = explode(':',
-                $redisConfig['services']['redis']['image']);
+            $imageRaw = $redisConfig['services']['redis']['image'] ?? '';
+
+            // Normalize image from possible environment syntax.
+            if (preg_match('/^\$\{[^:}]+:-(.+)\}$/', $imageRaw, $matches)) {
```
deviantintegral, line 474: Let's add a comment with an example of a string here?

src/ScaffoldInstallerPlugin.php
```diff
@@ -468,10 +468,26 @@ function($key) {
...
+                $image = $matches[1];
+            } elseif (preg_match('/^([^:]+):\$\{[^:}]+:-(.+)\}$/', $imageRaw, $matches)) {
+                $image = "{$matches[1]}:{$matches[2]}";
+            } else {
+                $image = $imageRaw;
+            }
+
+            // Extract the version/tag from the image string.
+            if (preg_match('/:([a-zA-Z0-9._-]+)$/', $image, $versionMatch)) {
```
deviantintegral, line 483: And an example here too!

src/ScaffoldInstallerPlugin.php
```diff
@@ -468,10 +468,26 @@ function($key) {
...
+            $imageRaw = $redisConfig['services']['redis']['image'] ?? '';
+
+            // Normalize image from possible environment syntax.
+            if (preg_match('/^\$\{[^:}]+:-(.+)\}$/', $imageRaw, $matches)) {
+                $image = $matches[1];
+            } elseif (preg_match('/^([^:]+):\$\{[^:}]+:-(.+)\}$/', $imageRaw, $matches)) {
+                $image = "{$matches[1]}:{$matches[2]}";
+            } else {
+                $image = $imageRaw;
+            }
```
deviantintegral, line 480: This could be simplified I think. If you change `$imageRaw = ` above to `$image`, then you can delete this else statement too?

deviantintegral (commented):
Do we also want to add a phpunit test for this?

src/ScaffoldInstallerPlugin.php
```diff
@@ -468,10 +468,26 @@ function($key) {
...
+            // Example: $image = 'redis:${REDIS_TAG:-6-bullseye}'.
+            elseif (preg_match('/^([^:]+):\$\{[^:}]+:-(.+)\}$/', $image, $matches)) {
+                $image = "{$matches[1]}:{$matches[2]}";
+            }
+        
+            // Extract the version/tag from the image string.
+            // Example: $image = 'redis:6-bullseye' → $version = '6-bullseye'.
+            if (!preg_match('/:([a-zA-Z0-9._-]+)$/', $image, $versionMatch)) {
+                throw new \RuntimeException("Unable to extract Redis version from image: {$image}");
+            }
```
mrdavidburns, line 487: Let's avoid code duplication and instead just call this new function.

`extractRedisImageVersion(string $image)`

mrdavidburns (approved):
LGTM! Thanks for the test!

Issue #347, opened 2023-12-06
Test failure shouldn't cause other tests to cancel execution


justafish:
https://github.com/go-task/task/issues/1246
https://github.com/go-task/task/blob/c4dce8f506814d7af750792f53d4fa0c863b4470/task.go#L291

elvism-lullabot:
@justafish @deviantintegral does this Issue have to do with the `cancel-in-progress` configuration?

justafish:
@elvism-lullabot no this is an issue with task. Say for example you wanted to run all your project's tests at once, a failure in one of the types of tests would halt execution in the others e.g. phpcs fails and kills all the other processes like PHPUnit

elvism-lullabot:
@justafish This is ready for review.

mrdavidburns:
This would ideally be fixed upstream in https://github.com/go-task/task/issues/1246

elvism-lullabot:
I’ve created a PR for the Task issue: https://github.com/go-task/task/pull/2307.
Next, I’ll look into how I can properly test it within this repository.

deviantintegral:
@rabbitlair could you test and review the upstream PR, and leave a comment there with anything you find?

rabbitlair:
As far as I can tell, everything works as described, and code changes make sense to me.

deviantintegral:
See https://github.com/go-task/task/issues/1246#issuecomment-3567079755 🥳

rabbitlair:
Task did published a new version that includes the changes required by this issue. I will create a pull request to upgrade Taskfile to latest.

PR #858, opened 2025-05-30
Issue #347: Fix test failure causes other tests to cancel.

elvism-lullabot:
see #347


tasks/test.yml
```diff
@@ -33,7 +57,6 @@ tasks:
             mkdir -p test_result
             ./vendor/bin/local-php-security-checker --format=json > test_result/local-php-security-checker.json
           fi
-        ignore_error: true
```
deviantintegral, line 36: I missed, why are we removing this?

tasks/test.yml
```diff
@@ -17,13 +17,37 @@ vars:
...
+        task test:phpstan
+        STATUS_PHPSTAN=$?
+        task test:phpunit:static
+        STATUS_PHPUNITSTATIC=$?
+        task test:phpcs
+        STATUS_PHPCS=$?
+
+        if [ $STATUS_LINT -ne 0 ] || [ $STATUS_PHPSTAN -ne 0 ] || [ $STATUS_PHPUNITSTATIC -ne 0 ] || [ $STATUS_PHPCS -ne 0 ]; then
+          exit 1
+        fi
```
deviantintegral, line 34: 🤔 this means these are all now run sequentially, and not in parallel. On the one hand, these are fast, on the other, parallel is even faster?

_however_, given this is an issue or feature request in Task itself, I'd love for us to instead fix or improve this upstream, as was linked in the original issue? See https://github.com/go-task/task/issues/1246

PR #861, opened 2025-06-02
Update Terminus to 4.0.1

plopesc:
Drainpipe currently uses Terminus 3.2.1, while the current version is 4.0.1. We should upgrade to the current version. This is a simple change in the following file: scaffold/gitlab/Pantheon.gitlab-ci.yml

The line:
curl -L https://github.com/pantheon-systems/terminus/releases/download/3.2.1/terminus.phar --output ~/terminus/terminus

should be changed to:
curl -L https://github.com/pantheon-systems/terminus/releases/download/4.0.1/terminus.phar --output ~/terminus/terminus

Old version also prevents the Drupal 11 migration since D11 requires PHP 8.3 and Terminus 3.2.1 is not fully compatible with PHP 8.3.


deviantintegral:
@plopesc I think Terminus 4 requires PHP 8.2+, which means we'd need to drop support for that in Drainpipe? Or download the right version based on the current PHP version being used?

plopesc:
Good point @deviantintegral

rabbitlair:
Closing as outdated - current Terminus version installed by Drainpipe is 4.1.1

PR #881, opened 2025-06-26
Issue #347: Use the binary that added failfast functionality.

elvism-lullabot:
see #347


deviantintegral:
Thanks for starting this. We probably want to be sure tests pass upstream before merging this!

rabbitlair:
The checks for _Test Binary Installer_ are failing due to the Taskfile version change, from `3.38` to `3.44.0-failfast-default-true`. Please, update the reference in [this file](https://github.com/Lullabot/drainpipe/blob/main/.github/workflows/TestBinaryInstaller.yml#L64) to fix it. Thanks!

rabbitlair (changes requested):
Please, see my comment above about changing the Taskfile version in the test action

PR #886, opened 2025-07-08
Issue #650: Fix drainpipe-dev tag push and allow phpunit v11 to fix composer conflicts

davereid:
Resolves https://github.com/Lullabot/drainpipe/issues/650

I have tested this with https://github.com/davereid/drainpipe and https://github.com/davereid/drainpipe-dev, which you can see was able to successfully push the latest tag release with this updated workflow:

- https://github.com/davereid/drainpipe-dev/releases/tag/v1.0.2
- https://github.com/davereid/drainpipe/actions/workflows/DrainpipeDev.yml
- https://github.com/davereid/drainpipe-dev/actions/workflows/release.yml

The problem was I was creating both a branch and a tag with the same name, then running `git push origin ${{ github.ref_name }}` and it didn't know to resolve the tag or the branch to push, so it failed.

This also gets rid of creating the release here in the workflow and letting the drainpipe-dev workflow create it when it sees a new tag.

### DIFF
--- a/.github/workflows/DrainpipeDev.yml
+++ b/.github/workflows/DrainpipeDev.yml
@@ -57,4 +57,4 @@ jobs:
           git reset --mixed origin/${{ github.ref_name }} || git reset --mixed origin/main
           git add -A
           git commit -m "${{ github.event.head_commit.message }}" --allow-empty
-          git push origin ${{ github.ref_name }}
+          git push origin refs/heads/${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


mrdavidburns:
@davereid Do we also need to make any changes to this commit?
https://github.com/Lullabot/drainpipe/commit/bab3cfc923f6bfaf36e704ae572350310adf716b

Our tests have been failing for new PRs and we suspect it's because of that change. 
It'll likely be why the tests on this PR also fail which will block us from being able to merge this.

davereid:
I think the actual issue was a composer conflict between drupal/core-recommended requiring sebastian/diff (~6.0.2) and our phpunit v10 dependency only allowing up to sebastian/diff ^5. By bumping the phpunit dependency up to v11 to match what core allows.

mrdavidburns (approved):
Looks good and thank you for fixing the phpunit to get tests passing.

PR #895, opened 2025-07-15
Update Tugboat configuration to capture mail

davereid:
The `settings.tugboat.php` file is not configured to capture emails by default, you have to configure this yourself. It would be nice if this could be done for us?

### DIFF
--- a/README.md
+++ b/README.md
@@ -789,6 +789,32 @@ solr:
   environment:
 ```
 
+### Mail Configuration
+
+By default, Drainpipe configures Tugboat environments to capture email using Tugboat's built-in mail capture service. This automatically configures:
+- The core mail system to use `php_mail`
+- The Mail System module (if installed) to use `php_mail` as the sender
+- Symfony Mailer (if installed) to send mail through Tugboat's SMTP service
+
+If you need to use a different mail configuration (e.g., for testing with a specific mail service or module), you can override these settings in your custom settings file (typically `web/sites/default/settings.tugboat.php` or `web/sites/default/settings.php`).
+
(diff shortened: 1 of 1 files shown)


deviantintegral:
I just went through some of this for a client project (but not in Tugboat). I discovered ddev does this by default:

```
// Override drupal/symfony_mailer default config to use Mailpit.
$config['symfony_mailer.settings']['default_transport'] = 'sendmail';
$config['symfony_mailer.mailer_transport.sendmail']['plugin'] = 'smtp';
$config['symfony_mailer.mailer_transport.sendmail']['configuration']['user'] = '';
$config['symfony_mailer.mailer_transport.sendmail']['configuration']['pass'] = '';
$config['symfony_mailer.mailer_transport.sendmail']['configuration']['host'] = 'localhost';
$config['symfony_mailer.mailer_transport.sendmail']['configuration']['port'] = '1025';
```

This actually led me to use the symfony mailer library instead of an alternative.

It strikes me that if this config is good enough for ddev, it's good enough for tugboat? I wonder if no matter what we do in Drainpipe if we should contribute a docs update to the tugboat pages with the above snippet.

mrdavidburns:
We discussed this again during the sync and agree that having this is a good default for all projects. 
We don't want projects that are using something that isn't `php_mail` (like MailGun or SendGrid) to be testing emails on Tugboat and accidentally have those test emails going out to real users.

What this PR needs before we can merge it is updated documentation in the README to let people know about this configuration and what they would need to do to override this in a custom settings file if they had a testing environment set up for something that isn't php_mail.

PR #901, opened 2025-07-24
Make Nightwatch.gitlab-ci.yml optional

plopesc:
Nightwatch.gitlab-ci.yml file is being created even if not desired.

We think that it might be created if only it's declared in the composer.json file under the extra.drainpipe.gitlab folder. like other scaffolding CI files.


m4olivei (approved):
I've run into this annoyance as well. It does seem like it should simply be optional. I've updated the `README` to surface it as an option. Would be nice to say something there about it, but I'm not familiar enough with how it actually gets used...

m4olivei:
Actually, @plopesc , I just realized that this is a duplicate: https://github.com/Lullabot/drainpipe/pull/706

Any preference for which to keep?

PR #915, opened 2025-08-15
drupal:update is missing one final cache:rebuild after deploy:hook

davereid:
As per https://architecture.lullabot.com/adr/20230929-drupal-build-steps/

```
vendor/bin/drush cache:rebuild --yes
vendor/bin/drush updatedb --yes
# Run config:import twice to make sure we catch any config that didn't declare
# a dependency correctly. This is also useful when importing large config sets
# as it can sometimes hit an out of memory error.
vendor/bin/drush config:import --yes || true
vendor/bin/drush config:import --yes
# Run update hooks that need to be run after config import. These need to be
# written as hook_deploy_NAME() hooks.
vendor/bin/drush deploy:hook --yes
vendor/bin/drush cache:rebuild -y
```

However we are missing that final `cache:rebuild` after running the `deploy:hook` command...

Also we should do a cache rebuild after turning off maintenance mode.

### DIFF
--- a/tasks/drupal.yml
+++ b/tasks/drupal.yml
@@ -66,7 +66,6 @@ tasks:
       # as it can sometimes hit an out of memory error.
       - ./vendor/bin/drush {{.site}} --yes config:import || true
       - ./vendor/bin/drush {{.site}} --yes config:import
-      - ./vendor/bin/drush {{.site}} --yes cache:rebuild
       - ./vendor/bin/drush {{.site}} --yes deploy:hook
       - ./vendor/bin/drush {{.site}} --yes cache:rebuild
       - |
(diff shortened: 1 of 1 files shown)


mrdavidburns (approved):

tasks/drupal.yml
```diff
@@ -86,3 +87,4 @@ tasks:
     desc: Turn off Maintenance Mode
     cmds:
       - ./vendor/bin/drush {{.site}} --yes state:set system.maintenance_mode 0 --input-format=integer
+      - ./vendor/bin/drush {{.site}} --yes cache:rebuild
```
deviantintegral, line 90: This shouldn't be needed, but also won't really hurt so 👍

deviantintegral (commented):
Thank you!

Issue #923, opened 2025-08-25
Investigate implementing Tugboat base preview to optimize build times and storage


mrdavidburns:
## Summary

We should investigate whether implementing [Tugboat's base preview feature](https://docs.tugboatqa.com/building-a-preview/work-with-base-previews/set-a-base-preview/index.html) could provide benefits for Drainpipe builds, particularly around storage optimization and potentially build time improvements.

## Background

Currently, each Drainpipe preview goes through a complete build process that includes:
- Installing system dependencies (Task, MariaDB client, Node.js, PHP extensions, etc.)
- Setting up the Drupal project structure 
- Installing Composer dependencies
- Configuring the environment

Even if Drainpipe requires a full rebuild for each pull request, there may still be significant value in using a base preview for the shared OS-level components and system setup.

### Storage Optimization
- **OS files sharing**: The base preview would contain all the system-level setup (PHP extensions, Node.js, system packages) that could be shared across all preview builds
- **Reduced disk usage**: Multiple previews would share the same base layer, reducing overall storage footprint
- **Faster image pulls**: Subsequent previews would only need to pull the delta changes rather than rebuilding everything from scratch

### Build Time Improvements
Even if we need to "wipe everything out" in terms of application code, we could still benefit from:
- Pre-installed system dependencies (Task, MariaDB client, Node.js, ImageMagick, etc.)
- Pre-configured PHP extensions (opcache, redis, gd)
- Pre-installed Composer and npm/yarn

### What could go in the base preview:
- [ ] System package installations (`apt-get` operations)
- [ ] PHP extension installations (opcache, redis, gd, ImageMagick)
- [ ] Node.js setup and corepack configuration
- [ ] Task installation
- [ ] Directory structure setup (potentially)
- [ ] Basic nginx/PHP-FPM configuration

### What would remain in preview builds:
- [ ] Git repository checkout
- [ ] Drupal project scaffolding (the `#drainpipe-start` sections)
- [ ] Composer install
- [ ] Application-specific configuration
- [ ] Database setup and imports

## Research Tasks

- [ ] **Analyze current build times**: Profile our existing build process to understand where time is spent
- [ ] **Review base preview documentation**: Understand Tugboat's base preview limitations and best practices
- [ ] **Identify optimization opportunities**: Determine which parts of our current init process are good candidates for the base preview
- [ ] **Test implementation**: Create a proof-of-concept base preview configuration
[...]

PR #941, opened 2025-09-15
Issue #923: Allow having a base preview for Drainpipe

rabbitlair:
This PR slightly modifies the Tugboat integration to allow having a base preview for Drainpipe repository. The approach is a bit different:

- Instead of creating a Drupal project in the `TUGBOAT_ROOT` directory, it is created in `/var/www/html`, which is a directory that already exists in the NGINX image.
- In the `TUGBOAT_ROOT` directory, the Drainpipe code from the corresponding branch/pull request is checked out and it is unaltered, so we do not need the logic to restore `.tugboat` directory after running Composer anymore.
- The `composer.json` file in the Drupal project is configured to install Drainpipe from the `TUGBOAT_ROOT` directory

The advantage of this approach is Tugboat can use the default behavior, which means we can create child previews as we would do with any other project.

Here you can see this PR in action - the base preview is 4.09 GB, but the child preview is 249.16 MB only:

image


.tugboat/steps/1-init.sh
```diff
@@ -63,6 +56,7 @@ EOD
 
 apt-get update
 apt-get -qq install nodejs
+apt-get clean
```
rabbitlair, line 59: This helps to reduce disk usage

.tugboat/steps/1-init.sh
```diff
@@ -72,24 +66,26 @@ nodejs -v | grep -q v$NODE_MAJOR
 #drainpipe-start
 # This is necessary for testing as this repository doesn't hold a Drupal site.
 shopt -s dotglob
-mkdir ../drainpipe-tmp
-mv * ../drainpipe-tmp/
-composer create-project drupal/recommended-project .
-mv ../drainpipe-tmp drainpipe
+rm -rf /var/www/html/*
+composer create-project drupal/recommended-project /var/www/html
```
rabbitlair, line 70: Drupal project is created in `/var/www/html` instead of in `TUGBOAT_ROOT`

.tugboat/steps/1-init.sh
```diff
@@ -72,24 +66,26 @@ nodejs -v | grep -q v$NODE_MAJOR
...
 
+# Create the Drupal private and public files directories if they aren't
+# already present.
+mkdir -p "${DOCROOT}/sites/default/files"
+chmod 777 "${DOCROOT}/sites/default/files"
+chgrp -R www-data "${DOCROOT}/sites/default/files"
+
 composer install
-#drainpipe-start
-rm -rf .tugboat
```
rabbitlair, line 93: This logic is not needed anymore, because Drainpipe is installed in `/var/www/html` and the directory `${TUGBOAT_ROOT}/.tugboat` remains unaltered.

PR #958, opened 2025-09-25
Add docs for act with OrbStack and Lima

deviantintegral:


deviantintegral:
Something is broken with lima. It's able to pull images and start the job, but then it changes permissions on `/var/run/docker.sock` inside of the lima VM and everything breaks.

```
andrew@lima-default:/Users/andrew$ ls -la /var/run/docker.sock
srw-rw---- 1 1001 rdma 0 Sep 25 10:01 /var/run/docker.sock
```

Issue #373, opened 2024-01-30
Create release guidelines


deviantintegral:
While we have a bunch of automated tests, we don't have current documentation for "things to test before rolling a release". There are some features that still need manual testing.

Let's create a document detailing how to roll a new Drainpipe release, including testing guidelines.

deviantintegral:
See also #188 .

rabbitlair:
The release guidelines will contain the following checks:
1. Drainpipe new version can be installed to a new Drupal project
2. Drainpipe new version can be updated from the current version in a Drupal project already using Drainpipe
3. All static and functional tests run properly
4. Deploys to Pantheon work (using Lullabot sandbox)
5. Deploys to Acquia work (using Lullabot sandbox)
6. Tugboat integration works

I will be automating as much checks as possible, and the base for testing will be the [Drainpipe Test](https://github.com/Lullabot/drainpipe-test) repository

See: https://github.com/Lullabot/drainpipe-test/issues/18

PR #1028, opened 2025-11-24
#373: Install a newer version of Terminus

rabbitlair:


deviantintegral:
@rabbitlair isn't terminus 4 out too?

rabbitlair:
There is a separate issue to upgrade Terminus to v4. I feel more comfortable tackling that one separate from the Pantheon deploy testing, so we can avoid potential issues of a major upgrade

Issue #1049, opened 2025-12-08
Fixing logic issue when defining/retrieving Node version


beto-aveiga:
### According to Copilot:
> The current command has a bug due to shell operator precedence. The pipe | has higher precedence than && and ||, so tr -d '[:space:]' is only applied when .nvmrc is not found. This can lead to NODE_MAJOR having trailing whitespace if .nvmrc exists and contains it.

There is a related Slack thread here: https://app.slack.com/client/T024KETL7/C08MT5PJAF5

PR #1050, opened 2025-12-08
#1049 Improving consistency in obtaining/determining Node version

beto-aveiga:
Closes: #1049 

image

The current command has a bug due to shell operator precedence. The pipe | has higher precedence than && and ||, so tr -d '[:space:]' is only applied when .nvmrc is not found. This can lead to NODE_MAJOR having trailing whitespace if .nvmrc exists and contains it.


rabbitlair (approved):
Looks good to me! Just a minor comment, but we can merge it without changing the code - I have no strong preference.

PR #1073, opened 2025-12-18
#263: Refactor Tugboat integration

### LINKED ISSUE #263: Thoughts on how we extend and customize the Tugboat integrations
I just went through adding the tugboat configs here to an existing project. Because it already had Tugboat configured, it gave me an opportunity to really see how far I could stretch things. Here's my notes from that process.

* It wasn't clear to me that checking for `sync:tugboat` was dynamic. I know the README calls it out, but perhaps composer install could log like "Checking for task sync:tugboat: not found, not added to Tugboat scripts".
* I think we should document how you can remove the composer config at https://github.com/Lullabot/drainpipe#tugboat after setup if you need to do significant overrides.

We needed to extend the tugboat config to add:

1. Visual diff URLs
2. A redis service (should this have come automatically?)
3. An additional dependency on a redis service
4. Aliases for multisites
5. A storybook service
6. Commands in mariadb to create additional databases and permissions for multisite databases
7. A command to clear the redis cache on start to ensure it's not stale in incorrect due to an unclean shutdown.

We ended up just not tracking upstream updates for now. But perhaps we could reuse the existing twig support to allow sites to customize the tugboat config? Also, I expected that task hooks were automatic, so that if we had a redis service, it would call `task tugboat:redis:build` and so on. It turns out that's hardcoded just for the PHP service.

rabbitlair:
Closes #263 

**Warning**: Contains a breaking change, so it requires, at least, a minor version bump.

This PR refactors the Tugboat integration to use DDEV configuration as the single source of truth for all service definitions. The new implementation generates a complete `config.yml` using Twig templates instead of separate bash scripts.

### Architecture
- **New dedicated plugin**: `TugboatConfigPlugin` handles all Tugboat logic (previously in `ScaffoldInstallerPlugin`)
- **Template-based generation**: Uses Twig template to generate `config.yml` file with sub-templates for each build phase in main PHP service
- **DDEV-driven configuration**: All service versions and types extracted from `.ddev/config.yaml` and docker-compose add-on files

### Build Phase Templates
- `config.yml.twig` - Main configuration template that includes phase templates
- `php-init.yml.twig` - System packages, PHP extensions, Node.js installation, composer install
- `php-update.yml.twig` - Composer install, sync task, file permissions changes, update tasks
- `php-build.yml.twig` - Build tasks

### Taskfile Integration
- **All services**: Supports `tugboat:{service}:init` tasks (e.g., `tugboat:redis:init`)
- **PHP service**: Supports `sync:tugboat`, `update:tugboat`, `build:tugboat`, `online:tugboat` tasks
- **Custom init script**: `custom-init-command.sh` wrapper for non-PHP service init tasks
- **Task installation**: Automatically installs Taskfile in init phase using version from `.taskfile` file

### Template Customization
- **Default templates**: Located in `vendor/lullabot/drainpipe/scaffold/tugboat/templates/`
- **Project overrides**: Place custom templates in `.tugboat/drainpipe-templates/`
- **Template precedence**: Project templates take priority over defaults
- **Custom Twig filters**: Added `indent` filter for proper YAML formatting

### Files Structure

```
scaffold/tugboat/
├── templates/
│   ├── config.yml.twig           # Main Tugboat config
│   ├── php-init.yml.twig         # Init phase commands
│   ├── php-update.yml.twig       # Update phase commands
│   ├── php-build.yml.twig        # Build phase commands
│   └── settings.tugboat.php.twig # Drupal settings
└── scripts/
    ├── custom-init-command.sh    # Init wrapper for non-PHP services
    └── install-mysql-client.sh   # MySQL client installer

.tugboat/ (generated)
├── config.yml                    # Generated config (overwritten on each run)
├── config.drainpipe-override.yml # User overrides (preserved)
├── drainpipe-templates/          # User template overrides (preserved)
└── scripts/                      # Copied scripts
```

Please, note this is a first working version. I would like to know your thoughts about the new approach - any feedback will be more than welcome. As far as I can tell, this refactor does not break backwards compatibility while adding the custom templates feature, which I think will make easier for users to customize their Tugboat integration. 

## Testing Instructions

First, create a new Drupal project and configure Drainpipe with Tugboat integration.

```
# Create a new Drupal project
mkdir testing
cd testing
composer create-project drupal/recommended-project . --ignore-platform-reqs
# Copy this Drainpipe branch to `testing/drainpipe` 

# Configure DDEV
ddev config --auto
ddev config --php-version "8.4"
ddev config --nodejs-version "24"
ddev get ddev/ddev-redis --version v1.2.0
ddev get ddev/ddev-elasticsearch --version v0.3.2
ddev start

# Configure Composer to work with Drainpipe
ddev exec --raw composer config extra.drupal-scaffold.gitignore true
ddev exec --raw composer config --json extra.drupal-scaffold.allowed-packages \[\"lullabot/drainpipe\"]
ddev exec --raw composer config --no-plugins allow-plugins.composer/installers true
ddev exec --raw composer config --no-plugins allow-plugins.drupal/core-composer-scaffold true
ddev exec --raw composer config --no-plugins allow-plugins.lullabot/drainpipe true
ddev exec --raw composer config repositories.drainpipe --json '{"type": "path", "url": "drainpipe", "options": {"symlink": false}}'
ddev exec --raw composer config minimum-stability dev

# Configure Tugboat integration
ddev exec --raw composer config extra.drainpipe --json '{"tugboat": {}}'
mkdir -p .tugboat
echo "php:" >  .tugboat/config.drainpipe-override.yml
echo "  aliases:" >> .tugboat/config.drainpipe-override.yml
echo "    - foo" >> .tugboat/config.drainpipe-override.yml
echo "  urls:" >> .tugboat/config.drainpipe-override.yml
echo "    - /" >> .tugboat/config.drainpipe-override.yml
echo "    - /?v=1" >> .tugboat/config.drainpipe-override.yml
echo "  screenshot:" >> .tugboat/config.drainpipe-override.yml
echo "    timeout: 45" >> .tugboat/config.drainpipe-override.yml
echo "  visualdiff:" >> .tugboat/config.drainpipe-override.yml
echo "    fullPage: false" >> .tugboat/config.drainpipe-override.yml

# Install Drainpipe
ddev composer require lullabot/drainpipe --with-all-dependencies
```

Verify the generated `.tugboat/config.yml` file is correct, and the services declared match the DDEV settings.

#### Test scenarios
1. Override a template for a PHP service's build phase
```
mkdir -p .tugboat/drainpipe-templates
cp vendor/lullabot/drainpipe/scaffold/tugboat/templates/php-update.yml.twig .tugboat/drainpipe-templates
# Add some new command to the file
ddev composer install
```
Then verify the new command is included in `.tugboat/config.yml`.

2. Add a custom `init` command for a secondary service:
```
echo "" >> Taskfile.yml
echo "  tugboat:redis:init:" >> Taskfile.yml
echo "    cmds:" >> Taskfile.yml
echo "      - whoami" >> Taskfile.yml
ddev composer install
```
Then verify the new command is included in `.tugboat/config.yml`.

3. Pantheon integration:
```
ddev exec --raw composer config extra.drainpipe.tugboat --json '{"pantheon": true}'
ddev composer install
```
Then verify Pantheon integration is included in `.tugboat/config.yml`.

#### More testing

Tested in a project currently using Drainpipe, so we can be confident about nothing breaking. [Tested on lullabotdotcom.](https://github.com/Lullabot/drainpipe/pull/1073#issuecomment-3725191115)

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -1,0 +1,13 @@
+have_fun: false
+memory_config:
+  disabled: false
+code_review:
+  disable: false
+  comment_severity_threshold: MEDIUM
+  max_review_comments: -1
+  pull_request_opened:
+    help: false
+    summary: false
+    code_review: true
+    include_drafts: true
+ignore_patterns: []
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -1,0 +1,84 @@
+# Drainpipe Code Review Style Guide
+
+## Review Tone and Format
+
+**Terse Communication**
+- Skip introductions, greetings, and sign-offs
+- Omit summary paragraphs unless specifically requested
+- Skip preambles like "I'm reviewing this pull request" or "Here's what I found"
+- Get directly to identified issues and recommendations
+
+**Eliminate Qualifiers**
+- Avoid hedging language: "might", "could", "possibly", "potentially", "seems like", "appears to"
+- Avoid softening phrases: "I think", "I believe", "in my opinion", "I would suggest"
+- State findings directly: Use "Missing error handling" instead of "It seems like error handling might be missing"
+- Be definitive when identifying issues
+
+**Minimize Positivity**
+- Skip positive feedback like "great", "excellent", "well-structured", "nice work"
+- Omit phrases like "this is a solid contribution" or "good job on..."
+- Focus exclusively on actionable items that need attention
+- If there are no issues, simply state "No issues found"
+
+## Review Structure
+
+**Format**
+- Use bullet points for multiple issues
+- One sentence per issue when possible
+- Group related issues together
+
+**Content Focus**
+- Report bugs, regressions, and breaking changes first
+- Highlight missing functionality or logic
+- Note security vulnerabilities
+- Identify performance issues
+- Point out maintainability concerns
+
+**Examples**
+
+BAD (too verbose, too positive):
+```
+Hello @user! I'm currently reviewing this pull request and will post my feedback shortly.
+This is a great refactoring that significantly improves the codebase. The new approach
+is well-structured and shows good understanding of the requirements. However, I did notice
+a few things that might need attention.
+```
+
+GOOD (terse, direct):
+```
+- Missing error handling in ConfigPlugin::generate()
+- Conditional logic for non-MariaDB databases removed from templates
+- Memcached configuration no longer generated for Apache setups
+```
+
+BAD (excessive qualifiers):
+```
+It seems like the error handling might be missing here. I think you should probably
+consider adding a try-catch block. This could potentially cause issues.
+```
+
+GOOD (direct):
+```
+Add error handling for file write operations.
+```
+
+## Code Review Categories
+
+Focus reviews on:
+- **Correctness**: Logic errors, regressions, missing conditions
+- **Security**: Vulnerabilities, insecure patterns
+- **Performance**: Bottlenecks, inefficient operations
+- **Maintainability**: Code clarity, duplication, complexity
+
+Omit commentary on:
+- Code style unless it impacts functionality
+- Minor formatting preferences
+- Subjective improvements that don't address issues
+
+## Response Guidelines
+
+- No introductory messages
+- No "I'll review this shortly" messages
+- No closing statements
+- Report only actionable findings
+- If no issues: remain silent or state "No issues found"
--- a/.github/workflows/TestTaskfileInstaller.yml
+++ b/.github/workflows/TestTaskfileInstaller.yml
@@ -15,10 +15,8 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        php-version: [ 8.1, 8.2, 8.3, 8.4, 8.5 ]
+        php-version: [ 8.2, 8.3, 8.4, 8.5 ]
         include:
-          - php-version: 8.1
-            drupal-version: ":^10"
           - php-version: 8.2
             drupal-version: ":^10"
           - php-version: 8.3
--- a/.github/workflows/composer-plugin.yml
+++ b/.github/workflows/composer-plugin.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        php-version: [ 8.1, 8.2, 8.3, 8.4, 8.5 ]
+        php-version: [ 8.2, 8.3, 8.4, 8.5 ]
     steps:
       - uses: actions/checkout@v6
 
--- a/.github/workflows/test-production-build.yml
+++ b/.github/workflows/test-production-build.yml
@@ -15,10 +15,8 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        php-version: [ 8.1, 8.2, 8.3, 8.4, 8.5 ]
+        php-version: [ 8.2, 8.3, 8.4, 8.5 ]
         include:
-          - php-version: 8.1
-            drupal-version: ":^10"
           - php-version: 8.2
             drupal-version: ":^10"
           - php-version: 8.3
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "psr-4": {"Lullabot\\Drainpipe\\Tests\\Functional\\": "tests/src/functional/"}
     },
     "require": {
-        "php": "^8.1",
+        "php": "^8.2",
         "ext-json": "*",
         "composer-plugin-api": "^2.0",
         "drush/drush": "^11|^12|^13",
@@ -20,7 +20,7 @@
     },
     "require-dev": {
         "composer/composer": "^2.9.3",
-        "phpunit/phpunit": "^9.6.21"
+        "phpunit/phpunit": "^9.6.31"
     },
     "extra": {
         "class": [
--- a/drainpipe-dev/.github/workflows/repo-lockdown.yml
+++ b/drainpipe-dev/.github/workflows/repo-lockdown.yml
@@ -16,7 +16,7 @@ jobs:
   action:
     runs-on: ubuntu-24.04
     steps:
-      - uses: dessant/repo-lockdown@v4
+      - uses: dessant/repo-lockdown@v5
         with:
           issue-comment: >
             Please file your issue in the upstream Drainpipe repository at https://github.com/Lullabot/drainpipe/issues.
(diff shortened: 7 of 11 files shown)


scaffold/tugboat/templates/config.yml.twig
```diff
@@ -0,0 +1,73 @@
...
+#   visualdiff:
+# solr:
+#   commands:
+#   checkout:
+#   depends:
+#   volumes:
+#   environment:
+services:
+  php:
+    http: false
```
deviantintegral: huh, should this be true?

rabbitlair:
@deviantintegral, Gemini comments were already addressed :)

Regarding test steps, that is a great idea. Let me create those, so this one is easier to review. Thanks!

rabbitlair:
I tested this PR in latest `main` from Lullabot D8 project, and it works like a charm. The only detail which is different is the calls in the build step are different. Currently, these are executed:
```
task build:tugboat
task drupal:update
task tugboat:drush-uli-ready
```
while the new branch replaces those to be:
```
task build
task update
task build:tugboat   # This is equivalent to tugboat:drush-uli-ready
```
We would need to update Lullabot-D8 to override the php-update Twig template, but apart from that... it should be good!

YesCT:
There's a note in the description:

> Also, take into account tests are broken - once we all are happy with the approach and implementation, I will fix those tests failing.

But I see 3 recent commits with passing tests green checkmarks. Hm. 

Are tests expected to be broken? Maybe that comment in the description is outdated. Or If it is still accurate, maybe it could be updated to list which tests. It's probably outdated. I'm gonna remove it. Put it back if I guessed wrong. ;)

.github/workflows/TestTugboat.yml
```diff
@@ -119,7 +110,6 @@ jobs:
         run: |
           ddev config --auto
           ddev config --php-version "8.3"
-          ddev config --nodejs-version "16"
```
YesCT: [Non-blocking Questions] 

why take this out?

cause node 16 is sooooo old we should be testing it?? ha :D 

do we test node override elsewhere?

why leave the php version config in??

hm.

.tugboat/config.yml
```diff
@@ -19,17 +20,89 @@ services:
...
+      - elasticsearch
     commands:
-      init: ./.tugboat/steps/1-init.sh
-      update: ./.tugboat/steps/2-update.sh
-      build: ./.tugboat/steps/3-build.sh
-      online: ./.tugboat/steps/4-online.sh
-
+      init:
+        - |
+          apt-get update
```
YesCT: hm. we loose the ability to run an automatic shell script syntax checker. 
why not keep the 1-init.sh pattern??

scaffold/tugboat/scripts/install-mysql-client.sh
YesCT: [Non-blocking Question] Is this an intended change?

If so, as part of review, should we check all the scripts (in scaffold/tugboat/scripts ??)are 644??

scaffold/tugboat/steps/1-init.sh.twig
YesCT: Just noting, we were doing twig templates before, this PR isn't introducing that.

src/TugboatConfigPlugin.php
```diff
@@ -0,0 +1,588 @@
...
+        // Example: ${REDIS_DOCKER_IMAGE:-redis:7}
+        if (preg_match('/^\$\{[^:}]+:-(.+)\}$/', $image, $matches)) {
+            $image = $matches[1];
+        }
+        // Example: redis:${REDIS_TAG:-7-alpine}
+        elseif (preg_match('/^([^:]+):\$\{[^:}]+:-(.+)\}$/', $image, $matches)) {
+            return $matches[2];
+        }
+
+        // Extract version from image:version format
```
YesCT: Thanks for the example format note.

src/TugboatConfigPlugin.php
```diff
@@ -0,0 +1,588 @@
...
+     * @return string The extracted version
+     */
+    private function extractVersionFromImage(string $image): string
+    {
+        // Handle environment variable syntax with fallback
+        // Example: ${REDIS_DOCKER_IMAGE:-redis:7}
+        if (preg_match('/^\$\{[^:}]+:-(.+)\}$/', $image, $matches)) {
+            $image = $matches[1];
+        }
+        // Example: redis:${REDIS_TAG:-7-alpine}
```
YesCT: <3

README.md
```diff
@@ -880,11 +923,8 @@ php:
   screenshot:
   visualdiff:
 solr:
-  commands:
   checkout:
   depends:
-  aliases:
-  urls:
```
YesCT: Are these changes unrelated to the refactor? Maybe removing invalid options that were invalid before even, could be it's own PR?? If so... [Request for changes]

.tugboat/config.yml
```diff
@@ -19,17 +20,89 @@ services:
...
+          apt-get install -y mariadb-client
```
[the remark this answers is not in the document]
rabbitlair: I tried to preserve all logic in the original shell scripts, but will double check in case I have missed something

scaffold/tugboat/scripts/install-mysql-client.sh
rabbitlair: Yes, it is an intended change - this script, and the other implemented in this refactor, are both scaffolded by Drainpipe, so I felt it was a good practice to remove the write permissions from them. I am open to keep the write permissions if we think that is better, of course

scaffold/tugboat/templates/settings.tugboat.php.twig
```diff
@@ -16,7 +16,7 @@ if (getenv('TUGBOAT_REPO')) {
     'username' => 'tugboat',
     'password' => 'tugboat',
     'prefix' => '',
-    'host' => '{{ database_type }}',
+    'host' => '{{ services.database.type }}',
```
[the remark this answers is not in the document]
rabbitlair: This change corresponds to the change on how the Twig variables are managed after the refactor. Instead of providing individual variables, everything related to Tugboat services is now passed as a single `services` array, hence the change from `database_type` to `services.database.type`.

.tugboat/config.yml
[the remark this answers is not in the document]
YesCT: [Very non-blocking] Would https://github.com/Lullabot/drainpipe/issues/771 have covered this?

rabbitlair:
After testing this branch with another real world project that has a complex Tugboat integration, I added the following changes:
* New template `php-online.yml.twig` to easily customize Online build phase
* Fixed a bug that prevented database service to properly run its custom init commands
* Fixed a bug when installing Taskfile to run custom init commands for all services
* Updated the MySQL repository GPG key, as the current one is expired
* Implemented support for using Percona as the Docker image for database service

YesCT (approved):

PR #1099, opened 2026-01-14
[DO NOT MERGE] Test Gemini behavior change

rabbitlair:
This pull request includes the Tugboat refactor for testing purposes only, and will be closed once Gemini provides its comments


rabbitlair:
Actually, I applied the suggestions from this test PR to the original Tugboat refactor, so I would say we are good to go!

PR #1104, opened 2026-01-20
chore: remove redundant composer install command from workflows

deviantintegral:


YesCT:
The commit message from the [commit](https://github.com/Lullabot/drainpipe/pull/1042/changes/bd4d495ed139b6390a988e6928e3fe5d4a33a3e0) that added the composer install said " Fix Composer Audit execution as Github Action" Maybe we should try and run the github action for composer audit? I don't see that action on this repo.. so maybe we need to try it on another repo (like a sandbox, or LDC) that is using drainpipe. cc @rabbitlair

rabbitlair:
I just verified that removing the first Composer run with the flag does not produce any issue: https://github.com/Lullabot/lullabot.com-d8/actions/runs/21185060429/

rabbitlair (approved):

PR #1125, opened 2026-02-19
1124: Silence SASS deprecation warnings (do not merge).

### LINKED ISSUE #1124: DEPRECATION WARNING: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.
This error occurs every time I run `ddev task sass`, once for every SCSS file:

```
DEPRECATION WARNING: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.

More info: https://sass-lang.com/d/legacy-js-api
```

Digging into it, it's caused by the `gulpfile.js` calling deprecated APIs.

quicksketch:
Fixes https://github.com/Lullabot/drainpipe/issues/1124

This is a temporary hack to silence the deprecation warnings when running `task sass`. It is not the correct solution but it shows where the problem lies in the current SASS compilation warnings.

### DIFF
--- a/metapackages/sass/gulpfile.js
+++ b/metapackages/sass/gulpfile.js
@@ -36,6 +36,7 @@ const includes = Object.keys(srcs)
     const result = dartSass.renderSync({
       file: file,
       includePaths: [modernNormalizePath],
+      silenceDeprecations: ['legacy-js-api'],
     });
     return result.stats.includedFiles.filter(file => typeof file === "string");
   })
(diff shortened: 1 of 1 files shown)


bronzehedwick:
@quicksketch Now that https://github.com/Lullabot/drainpipe/pull/1178 is merged, can this PR be closed?

quicksketch:
Should be able to, yes! I think we were applying this PR as a patch to Manhattan U. If we've updated Drainpipe and removed the patch, this should no longer be necessary.

PR #1149, opened 2026-03-05
feat: #728 Add Bitbucket Pipelines support for Acquia

### LINKED ISSUE #728: Set up Bitbucket Pipeline for Acquia deployment
depends on: #454 

Using the `task` commands set up in #454 let's add a composer.json Drainpipe configuration that will scaffold a Bitbucket Pipeline for deploying to Acquia. Let's have it so that this happens anytime something gets merged into `main` branch.

This would match functionality that we do with [PantheonReviewApps](https://github.com/Lullabot/drainpipe/blob/acdb3b4bb5aeea90058a19af76a350b58b33b019/scaffold/github/workflows/PantheonReviewAppsDDEV.yml)

rabbitlair:
Closes #728 

Adds Bitbucket Pipelines equivalents of the existing Acquia GitHub Actions workflows. Two pipelines are available: AcquiaReviewApps, which creates an Acquia Cloud Development Environment (CDE) for every pull request and posts the environment URL as a Bitbucket commit status; and AcquiaDeploy, which deploys to the Acquia dev environment whenever a PR is merged to the main branch.

Both pipelines are opt-in via composer.json and scaffold a bitbucket-pipelines.yml into the project on every composer install, kept up to date automatically. Pipeline logic is implemented as Taskfile tasks (tasks/bitbucket.yml) and invoked via ./vendor/bin/task, mirroring how the equivalent GitHub Actions workflows operate. User-defined Taskfile hooks (acquia:deploy:before, acquia:deploy:after) are supported, with sensible fallbacks when they are not defined. A scheduled cleanup pipeline is also included to remove stale CDEs that are no longer associated with an open pull request.

### Testing information

The `AcquiaDeploy` pipeline has been tested properly from a Bitbucket repository to an Acquia sandbox project:
image

Regarding `AcquiaReviewApp` pipeline, it should be working properly, but I have not fully tested because the Acquia sandbox where I tested does not allow CDEs:
image


README.md
```diff
@@ -797,6 +797,160 @@ include which will give you helpers that you can include and reference for tasks
...
 [scaffold/gitlab/Pantheon.gitlab-ci.yml](scaffold/gitlab/Pantheon.gitlab-ci.yml).
 
+## Bitbucket Pipelines Integration
+
+Add the following to `composer.json` to enable Bitbucket Pipelines support:
+
+```json
+"extra": {
+  "drainpipe": {
+    "bitbucket": []
```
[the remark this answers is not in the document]
beto-aveiga: Ok! No worries. It is mentioned/defined below.

deviantintegral (commented):
I haven't / can't easily test this. I left a few questions!

scaffold/bitbucket/setup-php.sh
```diff
@@ -0,0 +1,54 @@
...
+MINIMAL=false
+for arg in "$@"; do
+  case "$arg" in
+    --minimal) MINIMAL=true ;;
+  esac
+done
+
+if ! command -v composer &>/dev/null; then
+  COMPOSER_VERSION=2.9.5
+  curl -sS "https://getcomposer.org/download/${COMPOSER_VERSION}/composer.phar" -o /usr/local/bin/composer
```
beto-aveiga, line 25: We may be using curl before installing it.
Installing curl happens a few lines after.

rabbitlair: Ah, good catch, thanks a lot. Fix incoming

rabbitlair: Ok, fix applied. Thanks again!

beto-aveiga:
Hey, @rabbitlair !

Finally, I was able to make a full deploy. I really like the idea of installing PHP extensions by looking at Composer's requirements. That is smart!

For the project I was working on, the biggest issue with deployment was not with this Drainpipe code; it was on Acquia's side. The project has a different path for settings, but the Acquia Recommended Settings package does not play well with having a custom config path (via `$drs_override_config_directories`), so it always ends up using `../config/default`, which wipes all site configuration and breaks the deployment.

However, thanks to AI, I made a patch to address that situation, using a global variable in my settings.php to communicate with the Acquia script and use my custom config path.

From the following diff, I think the `acquia.settings.php` file needs a few changes that are probably not this PR's fault (see https://github.com/Lullabot/drainpipe/issues/1227).

What I modified was `.bitbucket/setup-php.sh` and the way `drupal:update` works, because `bitbucket:acquia:update-dev` was running locally in Bitbucket instead of remotely in Acquia DEV. 

Please let me know if you have any questions or comments.

[diff.txt](https://github.com/user-attachments/files/27278840/diff.txt)

rabbitlair:
I added a couple of commits to include the improvements to the `setup-php.sh` script, and added a new task to `tasks/bitbucket.yml` to ensure update commands are run in Acquia and not locally in the Bitbucket runner. Please, let me know how they look to you

beto-aveiga (approved):
Great work! Tested with a real site ✔️ 

Image

rabbitlair:
Awesome, thank you so much for all your testing, @beto-aveiga

PR #1191, opened 2026-04-03
feat!: #1173 Drop Nightwatch support

rabbitlair:
Closes #1173 

Drops Nightwatch support, keeping the related tasks as no-op displaying a warning - this is to prevent users currently using Nightwatch from having errors.


deviantintegral (commented):
How do you want to handle deprecating this? We could do a last 4.x release that shows a warning (so folks can get other changes), and then do a 5.x release? Other ideas?

PR #1204, opened 2026-04-10
fix: #766 support ed25519 keys

### LINKED ISSUE #766: Support ed25519 keys
https://github.com/Lullabot/drainpipe/pull/747/files#r1922471404

rabbitlair:
Closes #766 

The DDEV SSH setup in both the GitHub Actions composite action and the GitLab CI include was hardcoding the SSH key filename as `id_rsa`, which breaks `ed25519` keys since OpenSSH probes identity files by name. This pull request writes the key to a neutral filename (`id_drainpipe`) and adds an explicit `IdentityFile ~/.ssh/id_drainpipe` directive to `~/.ssh/config`, making the approach key-type-agnostic.

### DIFF
--- a/scaffold/github/actions/common/ddev/action.yml
+++ b/scaffold/github/actions/common/ddev/action.yml
@@ -17,59 +17,46 @@ inputs:
     description: "Composer cache directory, relative to the project workspace. Set to false to disable."
     required: false
   version:
-    description: "Override the DDEV version .e.g. '1.19.0'"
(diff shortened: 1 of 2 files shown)


deviantintegral (approved):
Obviously untested, but looks good to me.

Are there parts of this whole action we could replace or upstream into https://github.com/ddev/github-action-setup-ddev ?

rabbitlair:
I think we can simplify our action by using the one you mention. I will work on that soon.

PR #1211, opened 2026-04-15
feat: #553 add AsyncDeploy option for Pantheon review apps via Quicksilver webhook

### LINKED ISSUE #553: Make PantheonReviewApps more efficient
Make PantheonReviewApps more efficient and use fewer github minutes.

Idea via Andrew

One micro-optimization would be to check to see if the time is spent running drush commands, or waiting for pantheon to sync. We could start the command but not wait, let it run async, and let quicksilver post back to github to start a followup job to run the drupal steps?

rabbitlair:
Closes #553 

### Summary

- **Async deploy via Quicksilver:** Splits the blocking Pantheon review app job into two separate CI jobs, eliminating up to 10 minutes of idle runner time per PR push. A Quicksilver PHP hook fires a webhook from Pantheon to GitHub/GitLab after code sync completes, triggering a post-deploy job that runs Drush updates and marks the GitHub Deployment as `success`.
- **Cache Terminus phar:** The `setup-terminus` action now caches the Terminus binary with `actions/cache`, skipping the download on subsequent runs.
- **`PANTHEON_SKIP_WIPE_MULTIDEV` for GitLab:** GitLab review app jobs now support `PANTHEON_SKIP_WIPE_MULTIDEV`, matching the existing GitHub behaviour. When set to `"true"` and the multidev already exists, the delete-and-recreate step is skipped.

### Async deploy flow

```
Job 1 (PR-triggered, ~2–3 min)
  → build → create multidev → git push → exit   [runner released]

[Pantheon syncs code internally]

[Quicksilver fires drainpipe_notify_github.php / drainpipe_notify_gitlab.php]
  → reads secrets from Pantheon Secrets Manager
  → POSTs repository_dispatch to GitHub  /  triggers pipeline on GitLab

Job 2 (webhook-triggered, ~2–3 min)
  → composer install → terminus aliases → drush update → Deployment: success
```

### New opt-in config

```json
// GitHub
"github": { "pantheon": ["ReviewApps", "AsyncDeploy"] }

// GitLab
"gitlab": { "pantheon": ["Deploy", "ReviewApps", "AsyncDeploy"] }
```

Running `composer install` after adding `AsyncDeploy` scaffolds:
- `.github/workflows/PantheonReviewApps.yml` — replaced with async variant
- `.github/workflows/PantheonReviewAppsPostDeploy.yml` — new post-deploy workflow
- `web/private/scripts/drainpipe_notify_github.php` — Quicksilver script (+ GitLab equivalents)

Manual steps required after scaffolding: register the `sync_code` Quicksilver hook in `pantheon.yml` and set two Pantheon secrets (`drainpipe_github_token`, `drainpipe_github_repo`). Full instructions in the README.

### ⚠️ GitHub checks behaviour change in async mode

In async mode, Job 2 is triggered by `repository_dispatch` and does **not** appear as a PR check. A failed Drush update will not block a merge even if the review app check is required in branch protection. Job 2's result is visible in the GitHub Deployments panel on the PR.

### Async deploy observability

The async deploy pipeline introduces a new class of failure that the original implementation left silent: once Job 1 exits, the GitHub Deployment stays `in_progress` indefinitely if the Pantheon sync fails, if Quicksilver's webhook call is rejected, or if the dispatched Job 2 never starts. To surface these failures, every deployment status update now carries a `log_url` pointing directly to the CI run that created it — the "View deployment" link in the PR Deployments panel always navigates to the relevant logs regardless of which stage the pipeline is in. When Job 2 starts, it refreshes the `in_progress` status with its own run URL, so the link stays current as the pipeline hands off between jobs. The Quicksilver scripts also now exit with code 1 on non-2xx HTTP responses, causing Pantheon to record the `sync_code` workflow as failed rather than silently succeeding when the webhook dispatch is rejected.

For failures that happen entirely outside GitHub's visibility — a Pantheon sync error or a Quicksilver script that fails before firing the dispatch — a new `PantheonReviewAppsWatchdog.yml` workflow runs on a 30-minute schedule. It scans all GitHub Deployments stuck in `in_progress` past a configurable threshold (default 20 minutes, overridable via `PANTHEON_ASYNC_WATCHDOG_THRESHOLD_MINUTES`), queries `terminus workflow:list` to retrieve the Pantheon sync outcome, and marks each stale deployment `error` with a diagnostic description identifying whether the Pantheon sync failed, or whether sync succeeded but the post-deploy job never started. The watchdog is scaffolded automatically alongside the other async files when `AsyncDeploy` is added to `composer.json`. Job 1 also now writes a step summary to the Actions job page with a direct link to the Pantheon dashboard workflow logs and the `terminus workflow:list` command, giving developers a clear next step even before the watchdog fires.

### Test plan

- [ ] Add `"AsyncDeploy"` to a test project's `composer.json` and run `composer install` — verify the three async files are scaffolded and `PantheonReviewApps.yml` contains `async: 'true'`
- [ ] Confirm that `ReviewApps` without `AsyncDeploy` is unaffected (blocking workflow unchanged, no post-deploy file)
- [ ] `TestPantheonAsyncDeploy.yml` CI workflow covers: scaffold file placement (both modes), bash async guards in `review/action.yml` and `update/action.yml`, and Quicksilver PHP script pattern filtering + secrets validation
- [ ] Verify Terminus caching works: second run of `setup-terminus` skips the `curl` download step
- [ ] Verify `PANTHEON_SKIP_WIPE_MULTIDEV=true` on GitLab skips multidev wipe on subsequent pushes


rabbitlair:
I have tested this one against a real Pantheon project, and have applied some minor changes that were required. It is now confirmed that it works as described.

metapackages/javascript/package.json
```diff
@@ -6,7 +6,7 @@
     "@yarnpkg/esbuild-plugin-pnp": "^3.0.0-rc.15",
     "chokidar": "^5.0.0",
     "esbuild": "^0.28.0",
-    "yargs": "^18.0.0"
+    "yargs": "^17.7.2"
   },
```
deviantintegral, line 10: Why is this being downgraded?

metapackages/sass/package.json
```diff
@@ -13,7 +13,7 @@
     "modern-normalize": "^3.0.1",
     "postcss": "^8.5.8",
     "sass": "^1.97.3",
-    "yargs": "^18.0.0"
+    "yargs": "^17.7.2"
```
deviantintegral: Same

scaffold/github/actions/common/setup-ssh/action.yml
```diff
@@ -10,10 +10,12 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - run: |
+    - run: | # zizmor: ignore[github-env]
```
deviantintegral: Can we add a comment as to why this is safe to ignore?

scaffold/github/actions/pantheon/review/action.yml
```diff
@@ -34,6 +34,10 @@ inputs:
     description: "Whether or not to skip wiping the multidev on each push. Defaults to false."
     required: false
     default: "false"
+  async:
+    description: "When 'true', skip the post-deploy step (terminus workflow:wait, drush update, deployment success). The Quicksilver async flow handles that in a separate job."
```
deviantintegral:
```suggestion
    description: "When 'true', run the post-deploy step (terminus workflow:wait, drush update, deployment success) using a Quicksilver webook that calls back to GitHub."
```

scaffold/github/actions/pantheon/setup-terminus/action.yml
```diff
@@ -10,15 +10,26 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - env:
+    - name: Cache Terminus
+      id: terminus-cache
+      uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
+      with:
+        path: ~/terminus/terminus
+        key: terminus-4.1.1
```
deviantintegral, line 18: Add a renovate line to keep this in sync for future updates?

scaffold/github/actions/pantheon/setup-terminus/action.yml
```diff
@@ -10,15 +10,26 @@ inputs:
...
+        TERMINUS_CACHE_HIT: ${{ steps.terminus-cache.outputs.cache-hit }}
       run: |
         echo -e "Host *.drush.in\n    StrictHostKeyChecking no\n    HostkeyAlgorithms +ssh-rsa\n    PubkeyAcceptedAlgorithms +ssh-rsa" >> ~/.ssh/config
-        mkdir ~/terminus
-        curl -L https://github.com/pantheon-systems/terminus/releases/download/4.1.1/terminus.phar --output ~/terminus/terminus
-        chmod +x ~/terminus/terminus
-        ln -s ~/terminus/terminus /usr/local/bin/terminus
+        mkdir -p ~/terminus
+        if [ "$TERMINUS_CACHE_HIT" != "true" ]; then
+          curl -L https://github.com/pantheon-systems/terminus/releases/download/4.1.1/terminus.phar --output ~/terminus/terminus
```
deviantintegral, line 29: Same: Add a renovate line to keep this in sync for future updates?

I also wonder if we should put this in `/usr/local/bin`.

scaffold/github/actions/pantheon/setup-terminus/action.yml
```diff
@@ -10,15 +10,26 @@ inputs:
...
         INPUT_TERMINUS_PLUGINS: ${{ inputs.terminus-plugins }}
+        TERMINUS_CACHE_HIT: ${{ steps.terminus-cache.outputs.cache-hit }}
       run: |
         echo -e "Host *.drush.in\n    StrictHostKeyChecking no\n    HostkeyAlgorithms +ssh-rsa\n    PubkeyAcceptedAlgorithms +ssh-rsa" >> ~/.ssh/config
-        mkdir ~/terminus
-        curl -L https://github.com/pantheon-systems/terminus/releases/download/4.1.1/terminus.phar --output ~/terminus/terminus
-        chmod +x ~/terminus/terminus
-        ln -s ~/terminus/terminus /usr/local/bin/terminus
+        mkdir -p ~/terminus
+        if [ "$TERMINUS_CACHE_HIT" != "true" ]; then
```
deviantintegral, line 28: Instead of a variable, how about just `if [ ! -x /usr/local/bin/terminus ];` ?

scaffold/github/workflows/AcquiaDeploy.yml
```diff
@@ -13,6 +13,8 @@ on:
 jobs:
   deploy-to-acquia:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
deviantintegral, line 17: Should this be a separate PR?

scaffold/github/workflows/PantheonReviewAppsAsync.yml
```diff
@@ -0,0 +1,81 @@
...
+
+on:
+  pull_request:
+    types: [labeled, synchronize]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+# You will also need to add the secrets used below to Dependabot secrets
```
deviantintegral, line 11: What about Renovate?

scaffold/github/workflows/PantheonReviewAppsAsync.yml
```diff
@@ -0,0 +1,81 @@
+name: "Pantheon Review Apps"
```
deviantintegral: Should we have a different name here or not?

scaffold/github/workflows/PantheonReviewAppsAsync.yml
```diff
@@ -0,0 +1,81 @@
...
+          terminus-token: ${{ secrets.TERMINUS_MACHINE_TOKEN || secrets.PANTHEON_TERMINUS_TOKEN }}
+          terminus-plugins: ${{ vars.TERMINUS_PLUGINS || secrets.TERMINUS_PLUGINS }}
+
+      - uses: php-actions/composer@8a65f0d3c6a1d17ca4800491a40b5756a4c164f3 # v6.1.2
+
+      - name: Build Project
+        env:
+          directory: /tmp/release
+        run: |
+          task build
```
deviantintegral, line 69: The tricky bit I see here is what if the runner needs additional system dependencies for the project to build, that users expect to already have because they put them in ddev.

scaffold/github/workflows/PantheonReviewAppsWatchdog.yml
```diff
@@ -0,0 +1,148 @@
...
+          terminus-token: ${{ secrets.TERMINUS_MACHINE_TOKEN || secrets.PANTHEON_TERMINUS_TOKEN }}
+
+      - name: Detect and resolve stale deployments
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SITE_NAME: ${{ vars.PANTHEON_SITE_NAME || secrets.PANTHEON_SITE_NAME }}
+          THRESHOLD_MINUTES: ${{ inputs.threshold_minutes || vars.PANTHEON_ASYNC_WATCHDOG_THRESHOLD_MINUTES || '20' }}
+          TERMINUS_AVAILABLE: ${{ steps.terminus-setup.outcome == 'success' }}
+        run: | # zizmor: ignore[github-env]
+          set -euo pipefail
```
deviantintegral: This is long enough I think it should be pulled out into a shell script or into task, perhaps even rewritten in PHP or JavaScript)?

scaffold/github/workflows/Security.yml
```diff
@@ -21,11 +21,15 @@ jobs:
       pull-requests: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
```
deviantintegral, line 26: Do these other test workflow changes belong in this PR? Here down to TestStatic.

scaffold/gitlab/PantheonReviewAppsAsyncJob.gitlab-ci.yml
```diff
@@ -0,0 +1,109 @@
...
+        terminus aliases --only $PANTHEON_SITE_NAME --yes
+        if [ "$PANTHEON_MULTIDEV_RUN_INSTALLER" == "true" ]; then
+          ./vendor/bin/drush @$PANTHEON_SITE_NAME.mr-$CI_MERGE_REQUEST_IID --yes site:install --existing-config
+        elif ./vendor/bin/task -l | grep '* update: '; then
+          ./vendor/bin/task site=@$PANTHEON_SITE_NAME.mr-$CI_MERGE_REQUEST_IID update
+        else
+          ./vendor/bin/task site=@$PANTHEON_SITE_NAME.mr-$CI_MERGE_REQUEST_IID drupal:update
+        fi
+      fi
+    - echo $CI_ENVIRONMENT_URL > environment_url.txt
```
deviantintegral, line 31: This feels like code that should be shared between all CI providers.

scaffold/gitlab/PantheonReviewAppsAsyncJob.gitlab-ci.yml
```diff
@@ -0,0 +1,109 @@
...
+      for MR in $(terminus multidev:list $PANTHEON_SITE_NAME --format json | jq -r .[].id | cut -d '-' -f2); do
+        # Check it's a number.
+        if [ "$MR" -eq "$MR" ]; then
+          STATE=$(curl -s --fail -H "PRIVATE-TOKEN: $GITLAB_ACCESS_TOKEN" "https://gitlab.com/api/v4/projects/$CI_PROJECT_ID/merge_requests/$MR" | jq -r '.state')
+          if [ "$STATE" != "opened" ]; then
+            echo "Deleting Pantheon Multidev for $MR"
+            terminus --yes multidev:delete $PANTHEON_SITE_NAME.mr-$MR
+          fi
+        fi
+      done
```
deviantintegral, line 106: Same?

scaffold/pantheon/quicksilver/drainpipe_notify_github.php
```diff
@@ -0,0 +1,64 @@
...
+$http_code = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+curl_close($ch);
+
+if ($response === false || $http_code === 0) {
+    echo "ERROR: curl connection failed for $github_repo dispatch (environment: $environment).\n";
+    exit(1);
+}
+
+if ($http_code < 200 || $http_code >= 300) {
+    echo "ERROR: GitHub dispatch failed — HTTP $http_code for $github_repo (environment: $environment). Response: $response\n";
```
deviantintegral: Does pantheon have a timeout on these? Should we add a retry with some sort of backoff?

metapackages/javascript/package.json
rabbitlair: It was an error when solving a merge conflict, it is now back to the original version

metapackages/sass/package.json
rabbitlair: It was an error when solving a merge conflict, it is now back to the original version

scaffold/github/actions/common/setup-ssh/action.yml
rabbitlair: Comment added

scaffold/github/actions/pantheon/review/action.yml
rabbitlair: Change applied

scaffold/github/actions/pantheon/setup-terminus/action.yml
rabbitlair: Done!

scaffold/github/actions/pantheon/setup-terminus/action.yml
rabbitlair: Done. 

Regarding putting the binary in `/usr/local/bin`, it would make the file not cacheable - the cache path must be under the home directory or the workspace and `/usr/local/bin` is not restorable by the cache action.

scaffold/github/workflows/AcquiaDeploy.yml
rabbitlair: I moved it to a different pull request, good catch

scaffold/github/workflows/PantheonReviewAppsAsync.yml
rabbitlair: I added `(async)` to new async workflows so they have a more descriptive name

scaffold/github/workflows/PantheonReviewAppsWatchdog.yml
rabbitlair: Done

scaffold/github/workflows/Security.yml
rabbitlair: These are moved to a different PR

scaffold/pantheon/quicksilver/drainpipe_notify_github.php
rabbitlair: This is now done. Both Quicksilver scripts now retry up to 3 times with exponential backoff (2 s, then 4 s). Each attempt has a 10 s curl timeout, so the worst-case total is under Pantheon's 120 s Quicksilver limit.

scaffold/github/actions/pantheon/setup-terminus/action.yml
rabbitlair: The symlink at `/usr/local/bin/terminus` doesn't exist yet at that point in the script, so that check would always download. `~/terminus/terminus` would work, but `TERMINUS_CACHE_HIT` is the idiomatic pattern for `actions/cache` and stays correct if the cache path ever changes.

scaffold/github/workflows/PantheonReviewAppsAsync.yml
rabbitlair: Renovate tracks by file name, so would not need any change to work as expected

scaffold/github/workflows/PantheonReviewAppsAsync.yml
rabbitlair: Now, there are two scaffold versions for the workflow - one with DDEV, and the other (this one) does not use DDEV. I added a note to README file, so users are aware that `task build` must be self-contained, in order to have all required dependencies.

rabbitlair:
The only piece of feedback to be addressed at this moment is to extract the logic from [scaffold/gitlab/PantheonReviewAppsAsyncJob.gitlab-ci.yml](https://github.com/Lullabot/drainpipe/pull/1211/files/d089dea3169e3b3e6cde2fc5398b54e1ce148962#diff-8434b5553116e3af50e0466e06aeb6e2bd8281e3ae6fce05d36bc01fbedac1fe), so it can be used in any provider

deviantintegral:
Zequi is "99%" sure this is working well, but we would also like to test it on a project. We're going to merge this in so we can test without flipping to a dev release, given this entire new feature is opt in.

PR #1225, opened 2026-04-29
fix: remove unused GitHub API token inputs from action.yml

davereid:
I keep seeing errors when using `./.github/actions/drainpipe/set-env` in my project.

image

I went to look at the action and I don't see any usages of `github-api-token` nor `github-api-token-username` can we just remove them?


rabbitlair (approved):

rabbitlair:
Good catch, these input variables are never used in the action. Thanks!

PR #1229, opened 2026-05-05
Reference class path with 'use'  for EnvironmentDetector

### LINKED ISSUE #1227: Check for `acquia/drupal-recommended-settings`
In [settings.acquia.php](https://github.com/Lullabot/drainpipe/blob/main/scaffold/acquia/settings.acquia.php), we use the `EnvironmentDetector`, which is from `acquia/drupal-recommended-settings`, which depends on `acquia/drupal-environment-detector`, and the issue is that we are assuming this class is available, instead of checking and telling the user in case it is not, to avoid confusion.

Plus, we need to prepend a `use` statement (in `settings.acquia.php`) to let PHP know where our EnvironmentDetector is being called from.

`use Acquia\Drupal\RecommendedSettings\Helpers\EnvironmentDetector;`

beto-aveiga:
Avoids a fatal error when scaffolding the `settings.acquia.php` file

### DIFF
--- a/scaffold/acquia/settings.acquia.php
+++ b/scaffold/acquia/settings.acquia.php
@@ -7,6 +7,8 @@
 
 declare(strict_types=1);
 
+use Acquia\Drupal\RecommendedSettings\Helpers\EnvironmentDetector;
+
 if (getenv('AH_SITE_ENVIRONMENT')) {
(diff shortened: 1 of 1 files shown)


rabbitlair (approved):
Looks good!

PR #1239, opened 2026-05-22
fix(tugboat): Update mail system overrides to support Symfony Mailer v2

davereid:
The settings.tugboat.php template is missing the update configuration overrides when using the https://www.drupal.org/project/symfony_mailer 2.x version which was released on 2026-05-02, so fairly recently. This also matches the override that DDEV already supports for this.


rabbitlair (approved):
Looks good to me, thanks a lot!

PR #1316, opened 2026-09-16
fix: run pantheon:drupal-update through drainpipe_exec.

### LINKED ISSUE #1315: Pantheon deploys fail with "task: command not found" on DDEV projects
## Summary

In 6.2.0, `scaffold/github/actions/pantheon/update/action.yml` was refactored to call the new `pantheon:drupal-update` task. The step sources `bash_aliases` but then invokes `task` directly instead of going through `drainpipe_exec`:

https://github.com/Lullabot/drainpipe/blob/v6.2.0/scaffold/github/actions/pantheon/update/action.yml#L27-L32

```yaml
run: |
  source .github/actions/drainpipe/set-env/bash_aliases
  task pantheon:drupal-update \
    site="$INPUT_SITE_NAME" \
    environment="$INPUT_ENVIRONMENT" \
    skip_workflow_wait="$INPUT_SKIP_WORKFLOW_WAIT" \
    run_installer="$INPUT_RUN_INSTALLER"
```

On projects that run their deploy through DDEV (`DRAINPIPE_DDEV=true`), `task` doesn't exist on the runner — `vendor/bin/task` is a symlink to `/usr/local/bin/task`, which only exists inside the web container. The step fails immediately:

```
/home/runner/work/_temp/35a6425f-….sh: line 2: task: command not found
Error: Process completed with exit code 127.
```

This is the same class of failure as #511, in the same file.

## Reproduction

Hit on lullabot.com when upgrading Drainpipe from `^6.1.2` to `^6.2.0`.

- Release: https://github.com/Lullabot/lullabot.com-d8/releases/tag/release--2026-09-16.01
- Failed Development deploy: https://github.com/Lullabot/lullabot.com-d8/actions/runs/35117560231
- Failed Test deploy: https://github.com/Lullabot/lullabot.com-d8/actions/runs/35117674723

Both runs fail on the same step, `Run Updates`, which is the one using  `./.github/actions/drainpipe/pantheon/update`. Every preceding step passes, including the code push to Pantheon (`pantheon/push`, which still goes through `drainpipe_exec`).

## Impact

The update step runs after the code has already been pushed to Pantheon. On a typical deploy workflow it also sits between `drupal:maintenance:on` and `drupal:maintenance:off`, so the failure leaves the environment with:

- new code deployed, but `updb` / `cim` / `cache:rebuild` never run
- the site stuck in maintenance mode

alanjrl:
Closes #1315

## What was broken

Pantheon deploys were failing on the **Run Updates** step with `task: command not found` (exit 127).

The cause: `pantheon/update` called `task` directly. Nothing puts `vendor/bin` on the runner's `PATH`, so the binary was never found — on DDEV projects *and* non-DDEV ones.

This is bad timing for a failure: by then the code is already pushed, the site is left in maintenance mode, and `updb`/`cim` never run.

Introduced in 6.2.0 by 3f2f46b, which dropped the `drainpipe_exec` wrapper while refactoring this file.

## The fix

Call the task the same way every other action already does — through `drainpipe_exec`, with the explicit vendor path:

```bash
drainpipe_exec "./vendor/bin/task pantheon:drupal-update site=\"$INPUT_SITE_NAME\" ..."
```

`pantheon/update` was the only action still calling `task` bare. `pantheon/review/action.yml:179` makes a near-identical call and has always worked — this just brings `update` back in line.

`./vendor/bin/task` is correct in both environments: `TaskfileInstallerPlugin` either symlinks it to `/usr/local/bin/task` (DDEV container) or downloads the binary there (everything else).

## Regression test

Added a guard to the `bash-guards` job that fails if any composite action calls `task` without going through `drainpipe_exec`. It catches this exact regression, so 6.2.0 can't repeat itself.

Scope is limited to `scaffold/github/actions/**/action.yml` on purpose — the bare `task` calls in `PantheonReviewApps*.yml` are intentional (those workflows are explicitly non-DDEV).


.github/workflows/TestPantheonAsyncDeploy.yml
YesCT, line 1: Ah, this test greps the code. Ok. Could be helpful.

I went looking for a test that would run a deploy, and I think this step in the release process might have been missed. https://github.com/Lullabot/drainpipe/blob/main/RELEASE.md#3-integration-testing cc @rabbitlair

rabbitlair (approved):
Looks good, thank you!

